PR #1, opened 2015-07-11
switch from 'conditions' to 'ensure' per 0.57

vito:
`conditions` will be removed soon; `ensure` is the way to go, and has the advantage of running even if something errors


jtarchie:
Pulled in with modifications 042a067e709251db1922ca46886cf6efba7d1fbb

PR #2, opened 2015-07-11
ensure lock is released if fetching sources fails

vito:
previously the lock would only be released after running the task; if for whatever reason fetching resources fails, the lock would still be held.

this change comes at the cost of additional nesting in pretty much every job though, which might not be worth it if fetching resources pretty much never fails (the lock could always just be manually released).

feel free to close if so; just wanted to point it out.


jtarchie:
Closing, for now...

PR #3, opened 2016-06-09
detect concourse url from .flyrc

o-orand:
Tiny step, to be able to run buildpacks-ci on another concourse (ie not on  https://buildpacks.ci.cf-app.com)

PR #4, opened 2016-06-13
detect concourse url from .flyrc

o-orand:
It also set concourse_url, so it is available in erb template.

PR on develop instead of master.
Close #3


RochesterinNYC:
Thanks!

PR #5, opened 2016-07-01
Rubygem mirror support

o-orand:
Support a rubygem mirror to avoid network issue.

### DIFF
--- a/scripts/binary-builder.sh
+++ b/scripts/binary-builder.sh
@@ -8,6 +8,10 @@ ls spec/**/*_spec.rb | split -l $(expr `ls spec/**/*_spec.rb | wc -l` / $TOTAL_G
 specs=$(cat $(ls x* | sed -n "$CURRENT_GROUP p"))
 
 gem install bundler --no-ri --no-rdoc
+if [ "${RUBYGEM_MIRROR}" != "none" ]
+then
+    bundle config mirror.https://rubygems.org ${RUBYGEM_MIRROR}
+fi
 bundle install -j4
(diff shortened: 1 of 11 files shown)


scripts/binary-builder.sh
```diff
@@ -8,6 +8,10 @@ ls spec/**/*_spec.rb | split -l $(expr `ls spec/**/*_spec.rb | wc -l` / $TOTAL_G
 specs=$(cat $(ls x* | sed -n "$CURRENT_GROUP p"))
 
 gem install bundler --no-ri --no-rdoc
+if [ "${RUBYGEM_MIRROR}" != "none" ]
```
RochesterinNYC: These lines/if checking that checks for a value of `RUBYGEM_MIRROR` are actually not necessary.

If no `RUBYGEM_MIRROR` is specified, the command executed will be `bundle config mirror.https://rubygems.org`, which will simply return the local or global bundle config value for that mirror. Ex:

```bash
$ bundle config --local mirror.https://rubygems.org
Settings for `mirror.https://rubygems.org` in order of priority. The top value will be used
Set for the current user (/Users/pivotal/.bundle/config): "https://rubygems.org"
```

RochesterinNYC:
Once those `RUBYGEM_MIRROR` if conditionals are removed, this PR looks good to me. :+1:

o-orand:
@RochesterinNYC Thanks for your comments. Done and I've also added mirror support to update-compile-extensions.

RochesterinNYC:
Thanks!

PR #10, opened 2016-08-03
 introduce custom domain name with default value to cf-app.com.

o-orand:
Ignore DOMAIN_NAME in credential filter.
Use domain name only in bosh-lite task.
Use fallback default value in scripts when not set by tasks.


RochesterinNYC:
@o-orand thanks for the pull request. Generally looks great.

I notice that you flip between specifying `cf-apps` and `cf-app` for the domain in some places. Could you fix that please?

bin/update-all-the-pipelines
```diff
@@ -5,6 +5,22 @@ require 'yaml'
...
+
+lpass_config= {}
+lpass_yaml_data=%x{lpass show #{LPASS_CONCOURSE_PRIVATE} --notes && lpass show #{LPASS_DEPLOYMENTS_BUILDPACKS} --notes}
+if $?.exitstatus != 0
+  puts "WARNING: ignoring lastpass config file. An error occured while processing #{LPASS_CONCOURSE_PRIVATE} and #{LPASS_DEPLOYMENTS_BUILDPACKS}"
+else
+  lpass_config = YAML.load(lpass_yaml_data)
+end
+
+full_config = public_config.merge(lpass_config)
```
RochesterinNYC: What is the purpose of this `full_config` variable and the merge? It looks like only `domain-name` is only fetched from this `full_config`, and `domain-name` is specified in the public config, so I don't see why the `public_config` needs to be merged with the `lpass_config`?

public-config.yml
```diff
@@ -1,3 +1,4 @@
 # This file holds public credentials (like github uri, github organization) to ease creation of a new pipeline instance.
 buildpacks-github-org: cloudfoundry
 buildpacks-binaries-s3-bucket: pivotal-buildpacks
+domain-name: cf-apps.com
```
RochesterinNYC: I think this should be `cf-app.com`?

bin/update-all-the-pipelines
o-orand: It possible to override value define in `public_config` with some defined in `lpass-config`. It's the same behavior already in place in [set-pipeline](https://github.com/cloudfoundry/buildpacks-ci/blob/master/bin/update-all-the-pipelines#L22-L23). Is it clearer ?

public-config.yml
o-orand: You're right !

o-orand:
@RochesterinNYC typo fixed. Thanks !

RochesterinNYC:
Thanks.

PR #12, opened 2016-08-03
add slack-notification resource_types when required

o-orand:
To reduce concourse pre-requisite setup,  it adds slack-notification resource_types


RochesterinNYC:
@o-orand thanks for the pull request. Looks good.

PR #14, opened 2016-08-04
handle github organization in <language> buildpack uri.

o-orand:
reuse existing properties "buildpacks-github-org".

Should be merged after #10


RochesterinNYC:
Thanks @o-orand ! Great work.

PR #15, opened 2016-08-04
support a prefix in pipeline name (default: empty)

o-orand:


RochesterinNYC:
Looks good. Thanks @o-orand

PR #17, opened 2016-08-05
add current-timezone parameter for cron resource

o-orand:


RochesterinNYC:
Great stuff as always. Thanks @o-orand

PR #20, opened 2016-08-05
support machete git repo customization

o-orand:

PR #21, opened 2016-08-05
Custom machete git repo

o-orand:
it also includes a `public-config` refresh (ie. adds existing public properties)


o-orand:
see discussion #20.

RochesterinNYC:
Thanks!

PR #22, opened 2016-08-05
use concourse built-in variable in slack notifications for binary-bui…

o-orand:
…lder pipeline

Use this format $ATC_EXTERNAL_URL/pipelines/$BUILD_PIPELINE_NAME/jobs/$BUILD_JOB_NAME/builds/$BUILD_NAME

See #18

PR #23, opened 2016-08-05
use concourse built-in variable in slack notifications for every pipe…

o-orand:
…line

If notification format is fine with #22, you can merge this one.


RochesterinNYC:
Thanks @o-orand !

PR #25, opened 2016-08-09
minimize domain_name usage

o-orand:
to keep naming consistent with other keys

### DIFF
--- a/pipelines/templates/bosh-lite-cf-edge.yml
+++ b/pipelines/templates/bosh-lite-cf-edge.yml
@@ -123,7 +123,7 @@ jobs:
             BOSH_TARGET: <%= full_deployment_name +  '.' + domain_name %>
             BOSH_LITE_NAME: {{deployment-name}}
             RUBYGEM_MIRROR: {{rubygem-mirror}}
-            DOMAIN_NAME: {{domain_name}}
+            DOMAIN_NAME: {{domain-name}}
           ensure:
             put: deployments-buildpacks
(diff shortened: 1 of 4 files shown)


pipelines/templates/bosh-lite-cf-edge.yml
```diff
@@ -123,7 +123,7 @@ jobs:
             BOSH_TARGET: <%= full_deployment_name +  '.' + domain_name %>
             BOSH_LITE_NAME: {{deployment-name}}
             RUBYGEM_MIRROR: {{rubygem-mirror}}
-            DOMAIN_NAME: {{domain_name}}
+            DOMAIN_NAME: {{domain-name}}
```
RochesterinNYC: So where is this `domain-name` variable within the scope of the pipeline coming from? I thought that the value of `full_config["domain-name"]` is present in the pipeline variable scope as `domain_name` and not as `domain-name`?

o-orand: In fact, both are available !
 - `domain-name` from concourse credentials (through public-credentials or may be overridden with lastpass)
 - `domain_name` from erb, (only in ruby templating), but using value from concourse credentials using `full_config["domain-name"]` mechanism

If you prefer, I can rename `domain-name`, and `domain_name`, to `domain` or to any other name without _ nor -.

RochesterinNYC:
Sounds good. Thanks for the PR.

PR #33, opened 2016-08-16
add DOMAIN_NAME to  tasks in binary-builder

o-orand:
tasks: update-dependency-in-manifests and rspec


RochesterinNYC:
Thanks @o-orand !

PR #35, opened 2016-09-01
add retries to binary-builder tasks and set  RUBYGEM_MIRROR params

o-orand:


RochesterinNYC:
@o-orand are you seeing flaky behavior/frequent false failures with the binary builder tests?

o-orand:
@RochesterinNYC, not during the test, but during test setup (ie. failure while downloading bundler from rubygem)

o-orand:
@RochesterinNYC, if you don't want to add retry, it's also possible to do something like this https://github.com/orange-cloudfoundry/buildpacks-ci/blob/29161f4a3f43dfac748a9510e8aba6d045a494fc/scripts/binary-builder-unit.sh#L6-L8 or both :smile:

RochesterinNYC:
Wouldn't it be better to use the `--retry` flag that bundler provides? http://bundler.io/v1.12/bundle_install.html

My concern with multiple attempts is that the integration specs take a while, and retrying them in the case of an actual failure/issue will sap a lot of time/worker CPU.

o-orand:
In fact it's failing during bundler download, so I don't think retry option is going to solve the issue :disappointed:. I'm going to remove "attempt" and only set the mirror on gem install, is it fine for you ?

tasks/run-buildpack-checksums-specs.yml
```diff
@@ -13,6 +13,10 @@ run:
     - |
       set -ex
       cd buildpack-checksums
+      if [ -n "${RUBYGEM_MIRROR}" ]; then
+        gem sources --clear-all --add ${RUBYGEM_MIRROR}
+      fi
+      gem install bundler --no-ri --no-rdoc
       gem install bundle
```
RochesterinNYC: Can you remove this `gem install bundle` line as you're now using `gem install bundler`?

tasks/run-buildpack-checksums-specs.yml
```diff
@@ -13,6 +13,10 @@ run:
     - |
       set -ex
       cd buildpack-checksums
+      if [ -n "${RUBYGEM_MIRROR}" ]; then
+        gem sources --clear-all --add ${RUBYGEM_MIRROR}
+      fi
+      gem install bundler --no-ri --no-rdoc
```
RochesterinNYC: Additionally, it'd be great to see the `--no-ri` and `--no-rdoc` flags used for the other instances of `gem install bundler` in the codebase.

o-orand: There is only a few `gem install bundler` call in the codebase, so I'll do it.
```
scripts/binary-builder-unit.sh
9:  gem install bundler --no-ri --no-rdoc

scripts/binary-builder-integration.sh
9:  gem install bundler --no-ri --no-rdoc

Dockerfile
57:RUN gem install bosh_cli bosh_cli_plugin_micro rake bundler

tasks/run-buildpack-checksums-specs.yml
19:      gem install bundler --no-ri --no-rdoc

tasks/build-buildpack-checksums-site.yml
21:        gem install bundler
```
Do I also need to update the dockerfile ?

tasks/run-buildpack-checksums-specs.yml
o-orand: Yes ! I should have remove it ealier... Sorry.

RochesterinNYC:
Thanks!

PR #40, opened 2017-04-03
Use an S3 resource instead of `push-binary`

alexjh:
* https://github.com/cloudfoundry/buildpacks-ci/blob/f3b13fb7706c08e0d5e02c09a921809cf085237f/tasks/push-binary/s3-dependency-uploader.rb#L41 doesn't check the return value of the `aws` command. The behaviour I observe is that concourse spins forever on this step.
* Using an S3 resource makes it easier to use custom AWS endpoints because the change can be made in the pipeline config rather than the code for `s3-dependency-uploader.rb`

I tried to do some digging in the git history, but I didn't see any reasons why this couldn't be an S3 resource. If this PR is generally acceptable, I can change the other sections using the `push-binary` task and remove those files.

### DIFF
--- a/pipelines/binary-builder.yml
+++ b/pipelines/binary-builder.yml
@@ -87,6 +87,18 @@ resources: #####################################################################
         - CHANGELOG
 <% end %>
 
+  ## S3 Buckets ##
+
+<% %w(bower godep hwc dotnet composer glide nginx node yarn).each do |dependency| %>
+  - name: <%= dependency %>-binary
+    type: s3
+    source:
+      bucket: {{buildpacks-binaries-s3-bucket}}
+      regexp: dependencies/<%= dependency %>/(.*)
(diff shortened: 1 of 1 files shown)


sclevine:
Thanks @alexjh -- the team will take a look at this soon.

alexjh:
Thanks!

I've since realized that `extract-dotnet-framework` uploads multiple files, so that job would not be changed, and the `push-binary` task would stay. The rest of the jobs can change though.

dgodd:
@alexjh The reason we didn't use the s3-resource before was we wanted to check if a file existed in the bucket before overriding it. Since then, we have added an md5 stamp to the filename, so this is no longer a concern.

We have merged your commit and used it as inspiration to change `extract-dotnet-framework` as well.

Thanks!

alexjh:
@dgodd Wow, great, thank you! You've saved me a bunch of work today. :+1:

PR #41, opened 2017-05-12
Move skip_download to get_params

alexjh:
According to the [concourse put docs](https://concourse.ci/put-step.html), `skip_download` should be in the `get_params` section.


Zanadar:
Thanks for filing this PR. We'll look into the issue and make necessary changes on our end. Closing for now.

PR #46, opened 2018-10-03
Modify YML config to build new extensions for PHP

dmikusa-pivotal:
These changes are to be paired with https://github.com/cloudfoundry/binary-builder/pull/40 which allows us to build several new extensions for use with PHP.


tylerphelan:
Merged manually, changed default branch to develop so we can merge this in the UI

PR #47, opened 2019-01-14
Add Oauth token support to depwatcher

dragonchaser:
Signed-off-by: Ettore Di Giacinto <edigiacinto@suse.com>

This PR allows to use OAUTH tokens with the depwatcher (like github authentication tokens) by setting the environment variable `OAUTH_AUTHORIZATION_TOKEN`.

We added this feature because we constantly got http error 403 from github because too many people/pipelines on our network were accessing it.


Zanadar:
@dragonchaser Thanks for the PR. We'll take a look and prioritize.

PR #59, opened 2019-10-28
Rewrite update-cnb-dependency task in go

mdelillo:
**NOTE:** this is not ready to be merged yet. The task is (mostly) rewritten, see the bottom of this comment for what is left.

We are trying to simplify the inputs and params that this task takes. One simplification is to no longer support the `keep_latest_released` REMOVAL_STRATEGY. Instead, we will keep the latest two versions of each dependency line in child CNBs and the latest single version of each dependency in parent CNBs. This means we don't need a REMOVAL_STRATEGY anymore, just a number of dependencies to keep. We also don't need the latest released version of the buildpack. VERSION_LINE and VERSION_LINE_TYPE are being merged since we can figure out what to do based on the VERSION_LINE_TYPE. The `nginx` VERSION_LINE_TYPE is not necessary, we just need to remove old versions by hand from the `buildpack.toml` when we deprecate version lines.

Most of the logic of the task has been split out into four public functions:
`AddNewDep` adds the new dependency (one copy for each stack) to the dependencies from the latest `buildpack.toml`. It will always add the new version, even if it is not the latest or if it already exists (i.e. if we want to rebuild it).
`RemoveOldDeps` takes the dependencies from the previous step and removes all but the latest `n` versions of the given version line.
`UpdateDeprecationDates` adds (or overwrite) the deprecation date for the version line if it is provided.
`GenerateCommitMessage` uses the original and updated dependencies to figure out what has changed and return a commit message.

By splitting this logic up into four separate functions that don't depend on each other, they become much easier to test. For example, we don't need to know how many things we're going to keep, or what the version line is, in order to add a new dependency. Similarly, we don't need to know what was just added in order to remove old dependencies.

Another decision we made was to split out dependencies that had multiple stacks into multiple copies, one for each stack. It made the `AddNewDep` and `RemoveOldDeps` functions much easier to reason about. It helps us avoid edge cases, such as adding a dependency for some stacks when it already exists with a subset or superset of those stacks. If we wanted to, we could collapse these stacks again before committing. This way we are able to keep the simplicity they provide to the add/remove functions but don't change the way we write the `buildpack.toml`.

There are a few integration tests for `main`. They test the dependency updating, deprecation date updating, and commit message for both a child CNB and parent CNB. Unfortunately it is difficult to set them up because of the way the task must run. We needed to create files/directories to mimic the resources that are provided and pass parameters as environment variables. The output of the task is also a directory that we need to assert about. Because of all of this complexity, we don't test a ton of cases at that level. We could potentially have `main` just load all of the yaml/toml/json files and pass them to some other class that then calls our public functions. This way that class is a bit easier to test (just pass it structs instead of create files).

There is still some complexity in `main` that it would be great to find a way to pull out into a separate function. There is a loop that iterates over all of the relevant json files for the new dependency in `public-buildpacks-ci-robots`. For each one, it figures out which stack(s) the dependency is for based on the filename (handles `any-stack` and deprecated stacks). Then it creates a `Dependency` object by combining info from a few different inputs (mostly the json file it just read). We could just pull this out into its own function as-is, but the function may be annoying to test. We would need to create files with specific names and contents for each unit test.
A workaround could be to create structs to hold the files' names and contents and pass that into the function. Maybe that's a weird interface for the function to have? But at least it would keep most of the logic out of `main`.

## Work left to do

1. There are a few special cases in the ruby task ([1](https://github.com/cloudfoundry/buildpacks-ci/blob/a1264fcc03932a4a5122dd6c12163ae5d4333eb2/tasks/update-cnb-dependency/run.rb#L113-L116), [2](https://github.com/cloudfoundry/buildpacks-ci/blob/a1264fcc03932a4a5122dd6c12163ae5d4333eb2/tasks/update-cnb-dependency/run.rb#L128-L135), [3](https://github.com/cloudfoundry/buildpacks-ci/blob/a1264fcc03932a4a5122dd6c12163ae5d4333eb2/tasks/update-cnb-dependency/run.rb#L137-L145)) that should probably be ported over. Let's make sure they're still required first.

2. We changed what inputs and params are required but haven't updated the `task.yml`, pipeline, or `dependency-builds.yml` config file. We should be careful not to break anything that the `update-buildpack-dependency` task depends on. Some things in particular:
* We no longed need `buildpacks-latest-released` as an input
* We pass in an additional input `config` which is a directory containing the `dependency-builds.yml` file to make testing easier. We may need a param to point to the exact path, because in our case the resource will be `buildpacks-ci` but it is in the `pipelines/config` directory. Alternatively, we can recreate that structure in our `testdata`.
* We can stop passing in `VERSION_LINE_TYPE` and `REMOVAL_STRATEGY`. Instead we should pass in `VERSIONS_TO_KEEP` (generally 1 if it a parent CNB and 2 if it is a child CNB).

3. We don't explicitly set the git user (like [here](https://github.com/cloudfoundry/buildpacks-ci/blob/a1264fcc03932a4a5122dd6c12163ae5d4333eb2/tasks/update-cnb-dependency/run.rb#L222-L223)). Maybe we don't need to if our CI image does it?

4. Consider collapsing the stacks for each dependency before committing as described above.

5. Consider extracting the rest of the complexity from `main` as described above.


tasks/update-cnb-dependency/merge_dependency_lists.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import (
+	"sort"
+)
+
+func MergeDependencyLists(existingDeps, newDeps []Dependency) ([]Dependency, error) {
```
Zanadar: Maybe a method on `[]Dependency`?

pipelines/config/dependency-builds.yml
```diff
@@ -85,6 +91,7 @@ dependencies:
           - line: 2.2.4X
           - line: 3.0.1X
         removal_strategy: keep_all
+    versions_to_keep: 99
```
Zanadar: This is basically `keep_all`. It'll be nice to have a yml templating setup where was can use a variable here in place of this magic number.

tasks/update-cnb-dependency/buildpack_toml.go
```diff
@@ -0,0 +1,48 @@
...
+	Dependencies               []Dependency
+	DependencyDeprecationDates []DependencyDeprecationDate `toml:"dependency_deprecation_dates"`
+}
+
+type Stack struct {
+	ID string
+}
+
+func (buildpackToml BuildpackToml) LoadExpandedDependencies() []Dependency {
+	return expandDependenciesForEachStack(buildpackToml.Metadata.Dependencies)
```
Zanadar: I think this can be a method on `Dependencies`, which would make the organization a touch clearer.

tasks/update-cnb-dependency/commit_artifacts.go
```diff
@@ -0,0 +1,24 @@
...
+	if commitMessage == "" {
+		return nil
+	}
+
+	output, err := exec.Command("git", "-C", "artifacts", "add", "buildpack.toml").CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "failed to add artifacts: %s", string(output))
+	}
+
+	output, err = exec.Command("git", "-C", "artifacts", "commit", "-m", commitMessage).CombinedOutput()
```
Zanadar: I think this is cool for now, but we should consider using https://github.com/libgit2/git2go or something like it if things get more complicated.

tasks/update-cnb-dependency/load_resources.go
```diff
@@ -0,0 +1,120 @@
...
+	IncludeTiny      []string          `yaml:"include_tiny_in_any_stack"`
+}
+
+type EnvVars struct {
+	DeprecationDate DependencyDeprecationDate
+	VersionLine     string
+	VersionsToKeep  int
+}
+
+func loadResources() (EnvVars, DependencyOrchestratorConfig, BuildpackToml, Dependency, BuildMetadata, error) {
```
Zanadar: This is an example of where some more composition could be useful. Like this could return a single config object (for example). `Resources () (UpdateConfig, error)`

```go
type UpdateConfig struct {
  Env: Envars,
  Orchestrator: DependencyOrchestratorConfig,
   etc...
}
```

I think this would make things a bit more managable?

tasks/update-cnb-dependency/deprecation_date.go
```diff
@@ -0,0 +1,86 @@
...
+		VersionLine: versionLine,
+		Match:       match,
+	}, nil
+}
+
+func (d DependencyDeprecationDate) isNull() bool {
+	return d.Date == time.Time{} || d.Link == "" || d.VersionLine == "latest"
+}
+
+func UpdateDeprecationDatesWithDependency(deprecationDates []DependencyDeprecationDate, dep DependencyDeprecationDate) ([]DependencyDeprecationDate, error) {
```
Zanadar: Could be
```go
func (d []DependencyDeprecationDate) Update(dep DependencyDeprecationDate)
```

pipelines/config/dependency-builds.yml
```diff
@@ -787,6 +787,11 @@ v3_dep_names:
   miniconda3: Miniconda
   bundler: Bundler
   ruby: Ruby
+include_tiny_in_any_stack:
+  - dep
```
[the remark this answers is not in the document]
dfreilich: This is the dep dependency

tasks/update-cnb-dependency/commit_artifacts.go
dfreilich: 👍

tasks/update-cnb-dependency/deprecation_date.go
dfreilich: 👍

pipelines/config/dependency-builds.yml
mdelillo: 👍

PR #60, opened 2019-11-06
Address issue with buildpack.toml casing

Zanadar:
- Ensure arbitrary metadata is properly handled

[#169553482]

### DIFF
--- a/tasks/update-cnb-dependency/dependencies.go
+++ b/tasks/update-cnb-dependency/dependencies.go
@@ -16,7 +16,7 @@ type Dependency struct {
 	Name         string   `toml:"name,omitempty"`
 	SHA256       string   `toml:"sha256"`
 	Source       string   `toml:"source,omitempty"`
-	SourceSHA256 string   `toml:"source_sha256,omitempty"`
+	SourceSHA256 string   `mapstructure:"source_sha256" toml:"source_sha256,omitempty"`
 	Stacks       []string `toml:"stacks"`
 	URI          string   `toml:"uri"`
 	Version      string   `toml:"version"`
--- a/tasks/update-cnb-dependency/main.go
+++ b/tasks/update-cnb-dependency/main.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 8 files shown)


tasks/update-cnb-dependency/dependencies.go
```diff
@@ -16,7 +16,7 @@ type Dependency struct {
 	Name         string   `toml:"name,omitempty"`
 	SHA256       string   `toml:"sha256"`
 	Source       string   `toml:"source,omitempty"`
-	SourceSHA256 string   `toml:"source_sha256,omitempty"`
+	SourceSHA256 string   `mapstructure:"source_sha256" toml:"source_sha256,omitempty"`
```
dfreilich: Why don't any other of the keys have a mapstructure link? We only need it if it's a non-intuitive match?

tasks/update-cnb-dependency/main_test.go
```diff
@@ -353,6 +366,10 @@ func testUpdateCNBDependencyTask(t *testing.T, when spec.G, it spec.S) {
 			var buildpackTOML BuildpackTOML
 			_, err = toml.DecodeFile(filepath.Join(outputDir, "buildpack.toml"), &buildpackTOML)
 			require.NoError(t, err)
+
```
dfreilich: nit: Not equal spacing in all the additions - either all of that line (`var deps Dependencies`) should have a line above them, or none of them

tasks/update-cnb-dependency/main_test.go
```diff
@@ -353,6 +366,10 @@ func testUpdateCNBDependencyTask(t *testing.T, when spec.G, it spec.S) {
...
+			var deps Dependencies
+			require.NoError(t, mapstructure.Decode(buildpackTOML.Metadata["dependencies"], &deps))
+
```
dfreilich: nit: As above, make the spacing uniform between them

tasks/update-cnb-dependency/testdata/updating-child-cnb/buildpack.toml
```diff
@@ -8,6 +8,8 @@ version = "{{.Version}}"
 [metadata]
 include_files = ["bin/build", "bin/detect", "buildpack.toml"]
 pre_package = "./scripts/build.sh"
+random = "random"
+
```
dfreilich: nit: remove

tasks/update-cnb-dependency/main.go
```diff
@@ -58,7 +59,12 @@ func updateCNBDependencies() error {
 		return errors.Wrap(err, "failed to construct list of dependencies to add")
 	}
 
-	originalDeps := config.BuildpackTOML.Metadata.Dependencies.ExpandByStack()
+	var deps Dependencies
+	err = mapstructure.WeakDecode(config.BuildpackTOML.Metadata[DependenciesKey], &deps)
```
dfreilich: Why are we only using WeakDecode here? Is it to convert the version into a string?

tasks/update-cnb-dependency/dependencies.go
mdelillo: Right. When decoding json, yaml, or mapstructure (but not toml), the key name of the struct has to be a case-insensitive match of the data you're decoding. Since there's an underscore in this case, you have to tell it explicitly what the kay name is.

https://play.golang.org/p/g7Bq-vL2L4n

PR #63, opened 2019-11-11
169680854 collapse dependencies

dfreilich:
This implements a CollapseWithStack() method to collapse the expanded dependencies, such that equal dependency objects with different stacks will be combined. This should reduce the size of our `buildpack.tomls`.

### DIFF
--- a/tasks/update-cnb-dependency/dependencies.go
+++ b/tasks/update-cnb-dependency/dependencies.go
@@ -1,12 +1,12 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"path/filepath"
 	"sort"
 
 	"github.com/blang/semver"
+	"github.com/pkg/errors"
 )
 
 type Dependencies []Dependency
@@ -22,7 +22,18 @@ type Dependency struct {
 	Version      string   `toml:"version"`
 }
 
-func (deps Dependencies) MergeWith(newDeps Dependencies) (Dependencies, error) {
+func (deps Dependencies) Update(flags Flags, dep Dependency, depsToAdd Dependencies) (Dependencies, error) {
+	originalDeps := deps.ExpandByStack()
+	updatedDeps := originalDeps.MergeWith(depsToAdd)
+	updatedDeps, err := updatedDeps.RemoveOldDeps(dep.ID, flags.versionLine, flags.versionsToKeep)
+	if err != nil {
+		return Dependencies{}, errors.Wrap(err, "failed to remove old dependencies")
+	}
+
+	return updatedDeps.CollapseByStack(), nil
+}
+
+func (deps Dependencies) MergeWith(newDeps Dependencies) Dependencies {
 	depsMap := map[string]Dependency{}
 
 	for _, dep := range deps {
@@ -38,7 +49,7 @@ func (deps Dependencies) MergeWith(newDeps Dependencies) (Dependencies, error) {
 	}
 
 	sort.Slice(allDeps, allDeps.sortDependencies())
-	return allDeps, nil
+	return allDeps
 }
 
 func (deps Dependencies) RemoveOldDeps(depID, versionLine string, keepN int) (Dependencies, error) {
@@ -82,7 +93,8 @@ func (deps Dependencies) CollapseByStack() Dependencies {
 	for _, dep := range deps {
 		key := makeKeyWithoutStack(dep)
 		if mapDep, exists := depsMap[key]; exists {
-			mapDep.Stacks = combineStacks(mapDep.Stacks, dep.Stacks)
+			//Every dependency will be expanded, and will only have 1 stack
+			mapDep.Stacks = append(mapDep.Stacks, dep.Stacks[0])
 			depsMap[key] = mapDep
 		} else {
 			depsMap[key] = dep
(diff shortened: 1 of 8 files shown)


tasks/update-cnb-dependency/dependencies.go
```diff
@@ -22,14 +22,25 @@ type Dependency struct {
 	Version      string   `toml:"version"`
 }
 
-func (deps Dependencies) MergeWith(newDeps Dependencies) (Dependencies, error) {
+func (deps Dependencies) Update(flags Flags, dep Dependency, depsToAdd Dependencies) (Dependencies, error) {
```
mdelillo: `Dependencies` now has five public methods but only one (`Update`) is used outside of the struct. It's also one of the only methods without tests. Ideally, all of the other methods would be private, but that would mean we'd have to combine all of the test cases which we were trying to avoid. Maybe this method could be put somewhere else and tested? Or the caller could call the functions on `Dependencies` themselves?

mdelillo: I'd pass the `versionLine` and `versionsToKeep` in directly instead of the `flags` struct, it makes it more clear exactly what this function needs in order to do its work.

tasks/update-cnb-dependency/generate_commit_message.go
```diff
@@ -7,40 +7,61 @@ import (
...
+	added, rebuilt, stacks := findAddedDeps(dep, oldDeps, newDeps)
+
+	removedDeps, stacks := findRemovedDeps(dep, oldDeps, newDeps, stacks)
+
+	return formCommitMessage(dep, added, rebuilt, removedDeps, stacks, storyID)
+}
+
+func findAddedDeps(dep Dependency, oldDeps, newDeps Dependencies) (bool, bool, map[string]bool) {
+	var added, rebuilt bool
+	// Use maps to ensure no duplicates
```
mdelillo: Could we choose a better name for this variable to remove the need for a comment? Perhaps something like `uniqueStacks`?

tasks/update-cnb-dependency/main.go
```diff
@@ -49,57 +50,50 @@ func updateCNBDependencies() error {
 		return errors.Wrap(err, "failed to load task resources")
 	}
 
-	deprecationDate, err := NewDependencyDeprecationDate(flags.deprecationDate, flags.deprecationLink, config.Dep.ID, flags.versionLine, flags.deprecationMatch)
-	if err != nil {
-		return errors.Wrap(err, "failed to create deprecation date")
-	}
-
+	// Load, update and save dependencies
```
mdelillo: Could we choose different names for these methods/variables to remove the need for this comment?

tasks/update-cnb-dependency/main.go
```diff
@@ -49,57 +50,50 @@ func updateCNBDependencies() error {
...
 	}
-	originalDeps := deps.ExpandByStack()
-	updatedDeps, err := originalDeps.MergeWith(depsToAdd)
+
+	updatedDeps, err := deps.Update(flags, config.Dep, depsToAdd)
 	if err != nil {
-		return errors.Wrap(err, "failed to merge dependency lists")
+		return errors.Wrap(err, "failed to add new dependencies to the dependencies list")
 	}
+	config.BuildpackTOML.SaveDependencies(updatedDeps)
```
mdelillo: Instead of `BuildpackTOML` having separate methods to load and save dependencies that are only used here, it could have a single method that does all of this.

tasks/update-cnb-dependency/main.go
```diff
@@ -49,57 +50,50 @@ func updateCNBDependencies() error {
...
+	deprecationDates, err := config.BuildpackTOML.DeprecationDates()
 	if err != nil {
 		return errors.Wrap(err, "failed to decode deprecation dates")
 	}
+
 	updatedDeprecationDates, err := deprecationDates.Update(deprecationDate)
 	if err != nil {
 		return errors.Wrap(err, "failed to update deprecation dates")
 	}
+	config.BuildpackTOML.SaveDeprecationDates(updatedDeprecationDates)
```
mdelillo: Same as above (a single method on `BuildpackTOML` to update deprecation dates instead of loading and saving).

tasks/update-cnb-dependency/main.go
```diff
@@ -49,57 +50,50 @@ func updateCNBDependencies() error {
...
 
-	config.BuildpackTOML.Metadata[DependenciesKey] = updatedDeps
-	if len(updatedDeprecationDates) > 0 {
-		config.BuildpackTOML.Metadata[DeprecationDatesKey] = updatedDeprecationDates
-	}
-	// onfig.BuildpackTOML.Metadata = zeroMetadata(config.BuildpackTOML.Metadata => see comment below
-
+	// config.BuildpackTOML.Metadata = zeroMetadata(config.BuildpackTOML.Metadata => see comment below
+	// Update and save Order
```
mdelillo: Could we choose different names for these methods/variables to remove the need for this comment?

tasks/update-cnb-dependency/dependencies.go
dfreilich: Added as params

dfreilich: I'm not sure if my creation of the `UpdateDependenciesWith()` method in buildpack_toml helped or not. I'd love to hear your thoughts

tasks/update-cnb-dependency/generate_commit_message.go
dfreilich: Changed

tasks/update-cnb-dependency/main.go
dfreilich: I moved these methods to buildpack_toml with a clearer method name, to remove the comments

tasks/update-cnb-dependency/main.go
dfreilich: ^^ 👍

tasks/update-cnb-dependency/main.go
dfreilich: 👍

tasks/update-cnb-dependency/main.go
dfreilich: 👍

PR #64, opened 2019-11-13
Retain optional field for buildpack order.groups

Zanadar:
- Adds test for golden file of metabuildpack generation (see: https://youtu.be/8hQG7QlcLBk?t=823)

[#169716321]


Zanadar:
> I'd prefer `correct_buildpack.toml` over `golden_buildpack.toml`, which is a bit confusing

https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files

@dfreilich

PR #68, opened 2020-05-01
Remove incorrect words from jruby dependencies

andrew-edgar:
Relating this this issue https://github.com/cloudfoundry/ruby-buildpack/issues/125

We need to remove the invalid file from all ruby dependencies as well as ruby.

Thanks,

Andrew

### DIFF
--- a/tasks/build-binary-new/builder.rb
+++ b/tasks/build-binary-new/builder.rb
@@ -714,8 +714,13 @@ class Builder
         raise "Unsupported jruby version line #{source_input.version}"
       end
 
+
       # Create a copy of the source_input to prevent mutating version for later use
       full_version = "#{source_input.version}-ruby-#{ruby_version}"
+
+      filename = "#{binary_builder.base_dir}/#{source_input_name}-#{full_version}-linux-x64.tgz"
+      Archive.strip_incorrect_words_yaml_from_tar(filename)
+
       binary_builder.build(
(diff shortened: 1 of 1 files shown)


tasks/build-binary-new/builder.rb
```diff
@@ -714,8 +714,13 @@ def execute(binary_builder, stack, source_input, build_input, build_output, arti
         raise "Unsupported jruby version line #{source_input.version}"
       end
 
+
       # Create a copy of the source_input to prevent mutating version for later use
       full_version = "#{source_input.version}-ruby-#{ruby_version}"
+
+      filename = "#{binary_builder.base_dir}/#{source_input_name}-#{full_version}-linux-x64.tgz"
```
dumez-k, line 721: instead of `source_input_name` should be `source_input.name`

tasks/build-binary-new/builder.rb
```diff
@@ -714,8 +714,13 @@ def execute(binary_builder, stack, source_input, build_input, build_output, arti
...
+      Archive.strip_incorrect_words_yaml_from_tar(filename)
```
dumez-k, line 722: The strip must occur after the function `binary_builder.build` is called on line 724, otherwise the .tgz does not exist yet.

andrew-edgar:
Thanks I have updated those 2 items.  Appreciate the help!

PR #72, opened 2021-01-05
Update OCI8 version for PHP8

jfhovinne:
See https://pecl.php.net/package-changelog.php?package=oci8


mdelillo:
Looks good, thanks for the contribution!

PR #86, opened 2021-08-25
Remove dotnet-runtime version 2.1.X

bhh1893:


fg-j:
it seems like this change has already been applied to the concourse pipeline. any reason not to merge?

PR #91, opened 2021-11-01
Check the compiled artifact binaries are correct before upload

sophiewigmore:
During [this run](https://buildpacks.ci.cf-app.com/teams/main/pipelines/r-buildpack/jobs/detect-new-version-and-upload-artifacts/builds/50) of the R buildpack `detect-new-version-and-upload-artifacts` Concourse job, it was discovered that the uncached buildpack was not compiled correctly. We are not sure how this happened, but we do not check that the buildpack zip archives look correct when they come out of the buildpack packager. As a result, we published a broken buildpack.

This adds a check in which we unzip both the cached and uncached buildpacks and double check that the finalize/supply binaries look correct.

Asking for a review on this change before I push to master. and fly these updates

### DIFF
--- a/pipelines/templates/buildpack.yml.erb
+++ b/pipelines/templates/buildpack.yml.erb
@@ -459,6 +459,11 @@ jobs: ##########################################################################
                 pivotal-buildpack-cached: pivotal-buildpack-cached-<%= stack %>
               output_mapping:
                 buildpack-artifacts: buildpack-artifacts-<%= stack %>
+            - task: check-built-artifacts
+              file: buildpacks-ci/tasks/check-built-artifacts/task.yml
+              input_mapping:
+                pivotal-buildpack: pivotal-buildpack-<%= stack %>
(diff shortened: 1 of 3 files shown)


sophiewigmore:
bump @arjun024

arjun024:
My concern is that, in this change a symptom of the problem is checked to verify the binaries. A wrong binary in the right format could easily pass the test and give us false confidence.

Buildpacks also need not have all these binaries (e.g. apt-buildpack has no finalize). And not sure if we have other buildpacks where bins are in some other executable format

sophiewigmore:
@arjun024 This is true. I'm having trouble thinking of an alternative we could do to check the artifact before it gets uploaded. Clearly after a full investigation of this problem, we couldn't pinpoint the actual problem. My thoughts were that some kind of check would be better than nothing. We could try adding a single buildpack test for the cached/uncached buildpacks?

sophiewigmore:
Closing this due to my own lack of activity

PR #98, opened 2022-01-10
Add missing pipenv dependency flit_core to build

thitch97:
# Background
Offline builds of applications which use pipenv were observed to fail with the following error:
```
ERROR: Could not find a version that satisfies the requirement flit_core<4,>=3.2.0 (from versions: none)
ERROR: No matching distribution found for flit_core<4,>=3.2.0
```
[Example Github Actions output](https://github.com/paketo-buildpacks/pipenv/runs/4767335181?check_suite_focus=true#step:5:398)

# Solution
The `flit_core` package appears to be an installation requirement for pipenv itself. Since it is not vendored as part of the pipenv source, pipenv attempts to download the missing artifacts during its installation which is not possible in an airgapped environment. This PR adds `flit_core` to the list of dependencies that are explicitly downloaded and included in our pipenv distribution.

### DIFF
--- a/tasks/build-binary-new/builder.rb
+++ b/tasks/build-binary-new/builder.rb
@@ -75,6 +75,7 @@ module DependencyBuild
           Runner.run('python3', '-m', 'pip', 'download', '--no-binary', ':all:', 'parver')
           Runner.run('python3', '-m', 'pip', 'download', '--no-binary', ':all:', 'wheel')
           Runner.run('python3', '-m', 'pip', 'download', '--no-binary', ':all:', 'invoke')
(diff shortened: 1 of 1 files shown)


menehune23 (approved):

menehune23:
@arjun024 @brayanhenao curious about your thoughts here.

I'm comfortable with this, given it's for pipenv itself vs some other dependency (which should just be vendored instead for offline environments).

brayanhenao (approved):
LGTM

arjun024 (approved):

PR #102, opened 2022-01-13
Update AppDynamics PHP agent files from DepWatcher

brayanhenao:


dockerfiles/depwatcher/src/depwatcher/app_dynamics_agent.cr
```diff
@@ -30,18 +31,18 @@ module Depwatcher
...
-          nil
-        end
-      end.compact.sort_by { |r| Version.new(r.ref) }.last(10)
+      allReleases = Array(Release).new
+      response = client.get("https://download.run.pivotal.io/appdynamics-php/index.yml").body
+      response.each_line do |appdVersion|
+        splitArray = appdVersion.split(": ")
+        version = splitArray[0].sub("_", "-")
+        url = splitArray[1]
+        File.write("#{version}.tar.bz2", client.get(url).body)
```
menehune23, line 40: Rather than assume `tar.bz2`, it would be best to use the file name present in the URL

menehune23: Actually, now that I think about it: since you're not doing any decompressing, you could choose any file name (or no extension even)

brayanhenao: Yes, It's only computing the SHA256, so any extension/name of the file should work.

dockerfiles/depwatcher/src/depwatcher/app_dynamics_agent.cr
```diff
@@ -66,8 +67,14 @@ module Depwatcher
...
 
+    # AppDynamics uses a different versioning scheme than SemVer. They use calendar versioning (ref https://community.appdynamics.com/t5/Knowledge-Base/New-in-March-2020-AppDynamics-is-switching-to-calendar/ta-p/38364)
+    # So every new release will follow the same pattern: YY.M.P-X  (YY = year, M = month, P = patch, X = metadata). Example: 22.1.0-14
     def initialize(@original : String)
-      @major, @minor, @patch, @metadata = @original.split(".").map(&.to_i)
+      splitVersion = @original.split(".")
+      @major = splitVersion[0].to_i
+      @minor = splitVersion[1].to_i
+      @patch = splitVersion[2].split("-")[0].to_i
+      @metadata = splitVersion[2].split("-")[1].to_i
```
menehune23, line 77: Consider storing `splitVersion[2].split("-")` rather than retyping, as it seems like something that could get out of sync easily

menehune23 (changes requested):
Overall looks great! Just a couple of minor comments

menehune23 (approved):
🚀

PR #104, opened 2022-01-31
Update the Python version used to build Pip dependency

brayanhenao:
# Context

With the release of `Pip 22.0.0` any version of Pip `≥22.0.0` will require a Python version `≥ 3.7` ([ref](https://github.com/pypa/pip/pull/10641)).

# Solution

- Removed the command `apt-get install python3-pip` since it installs `Python 3.6.8`.
- Add the command `apt install` to install `python3.7` and `python3.7-distutils`.
- Install pip tool using the [`get-pip.py`](https://bootstrap.pypa.io/get-pip.py) script

### DIFF
--- a/tasks/build-binary-new/builder.rb
+++ b/tasks/build-binary-new/builder.rb
@@ -18,29 +18,18 @@ end
 
 module DependencyBuild
   class << self
-
-    def self.build_python_for_pip(python_version)
-     # The python_version argument is the version of python that we are building for pip. It requires major, minor and
-     # patch. Eg: 3.7.4
-
-     Runner.run('apt', 'update')
-     Runner.run('apt', 'install', '-y', 'build-essential', 'zlib1g-dev', 'libncurses5-dev', 'libgdbm-dev', 'libnss3-dev', 'libssl-dev', 'libsqlite3-dev', 'libreadline-dev', 'libffi-dev', 'curl' ,'libbz2-dev')
-     Runner.run('mkdir', '-p', '/tmp/python/')
-     Runner.run('curl', '-L', 'https://www.python.org/ftp/python/' + python_version + '/Python-' + python_version + '.tgz', '-o', '/tmp/python/Python-' + python_version + '.tgz')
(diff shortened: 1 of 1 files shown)


arjun024:
> With the release of Pip 20.0.0 any version of Pip ≥20.0.0
Do you mean 22? Isn't 20 a two year old version?

tasks/build-binary-new/builder.rb
```diff
@@ -18,12 +18,29 @@ def run(*args)
...
       # final resting place for pip source and dependencies
       file_path = "/tmp/pip-#{source_input.version}.tgz"
       ENV['LC_CTYPE'] = 'en_US.UTF-8'
-      Runner.run('apt', 'update')
-      Runner.run('apt-get', 'install', '-y', 'python3-pip')
+
+      # For the latest version of pip, it requires python version >= 3.7 (ref: https://github.com/pypa/pip/pull/10641),
+      # so we need to build python >= 3.7 first.
+
+      build_python_for_pip('3.7.11')
```
arjun024, line 42: I'm concerned that hardcoding version to a patch would mean that we have a high chance of missing out on critical security fixed.
Any reason we can't replace the function with something like the following example?
```bash
apt update
# or any 3.x version you like
apt-get install -y python3.8
# python3.8 --version
# Python 3.8.0
```

brayanhenao: When using `apt install python3.7` pip is not installed, so it must be configured manually using [this file](https://bootstrap.pypa.io/get-pip.py). This requires another package called `python3.7-distutils` to be installed.

Do you think it is clearer to install Python/Pip using this method?

brayanhenao: @arjun024 PR Updated!

menehune23:
Anyone on Paketo we should tag in this PR?

menehune23 (approved):

brayanhenao:
C.C: @sophiewigmore

[Pip Build and Upload](https://github.com/paketo-buildpacks/dep-server/actions/workflows/pip-build-upload.yml) has been failing for the same reason, this should fix it.

PR #111, opened 2022-03-08
Update .NET Buildpack dependencies lines to only keep 1 of each patch version (latest)

brayanhenao:
Related to https://github.com/cloudfoundry/dotnet-core-buildpack/issues/518.

### DIFF
--- a/pipelines/config/dependency-builds.yml
+++ b/pipelines/config/dependency-builds.yml
@@ -59,26 +59,14 @@ dependencies:
           - line: 3.1.4X
             deprecation_date: 2022-12-03
             link: https://dotnet.microsoft.com/platform/support/policy/dotnet-core
-          - line: 5.0.1X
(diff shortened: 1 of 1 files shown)


arjun024:
LGTM.
My only comment is that `versions_to_keep` seems to be irrelevant and be only used for updating dependencies in CNBs which we don't do anymore - but that's to clean up for a later task.

I'm not merging this just to make sure that you don't forget to `fly` the pipeline. You can merge it after you fly.

arjun024:
LGTM.
My only comment is that `versions_to_keep` seems to be irrelevant and be only used for updating dependencies in CNBs which we don't do anymore - but that's to clean up for a later task.

I'm not merging this just to make sure that you don't forget to `fly` the pipeline. You can merge it after you fly.

arjun024 (approved):

PR #151, opened 2022-07-19
Cflinuxfs4 dependency builds refactor

brayanhenao:
# Changes

- Move `create-new-version-line-story` tasks from `tasks/build-binary-new/` to `tasks/create-new-version-line-story/` (Separation of responsibilities). 
- Create new folder `tasks/build-binary-new-cflinuxfs4`: Since the version of ruby used in `cflinuxfs4` is `3.0.x` the code requires some updated. It was initially designed for `Ruby 2.x`. In a near future (next year) when `cflinuxfs3` is deprecated we can rename this folder and remove the old one with dead `Ruby 2.x` code.
- All calls to the [`binary-builder`](https://github.com/cloudfoundry/binary-builder) were updated to reflect the new folder containing all the new code used for build deps in `cflinuxfs4`. See https://github.com/cloudfoundry/binary-builder/pull/58 

# Note

ONLY Merge this after https://github.com/cloudfoundry/binary-builder/pull/58 is merged.


pipelines/config/dependency-builds-cflinuxfs4-test.yml
```diff
@@ -0,0 +1,104 @@
+---
```
arjun024, line 1: did you mean to check-in this file, or was this just for temporary testing?

pipelines/config/dependency-builds.yml
```diff
@@ -249,6 +249,7 @@ dependencies:
     source_type: node
     copy-stacks:
       - bionic
+      - jammy
```
arjun024, line 252: Do we need the stack called `jammy`? I believe the copy to `bionic` stack was introduced to support paketo at some point (who don't use this pipeline anymore).
Question applies to all places with jammy reference

pipelines/dependency-builds-cflinuxfs4-test.yml.erb
```diff
@@ -0,0 +1,297 @@
+<%
```
arjun024, line 1: Same question as above - was the test pipeline intended to be checked-in?

pipelines/dependency-builds.yml.erb
```diff
@@ -168,6 +169,9 @@ resources:
 <% stacks.each do |stack| %>
 - name: <%= stack %>-image
   type: docker-image
+  <% if stack == 'cflinuxfs4' %>
+  source: {repository: cfbuildpacks/<%= stack %>}
+  <% end %>
   source: {repository: cloudfoundry/<%= stack %>}
```
arjun024, line 175: do we need an `else` here?

pipelines/config/dependency-builds.yml
```diff
@@ -381,8 +382,9 @@ dependencies:
     versions_to_keep: 2
     copy-stacks:
       - bionic
-build_stacks: [ 'cflinuxfs3' ]
-copy_stacks: [ 'bionic' ]
+cflinuxfs4_dependencies: [ 'node' ]
```
arjun024, line 385: Are these all the dependencies that will be built against cflinuxfs4?
In that case, to support dotnet-core buildpack, shouldn't this also include `libgdiplus` and `libunwind`?

pipelines/config/dependency-builds.yml
```diff
@@ -381,8 +382,9 @@ dependencies:
...
+build_stacks: [ 'cflinuxfs4' , 'cflinuxfs3' ]
+copy_stacks: [ 'jammy', 'bionic' ]
 windows_stacks: [ 'windows2016', 'windows' ]
 deprecated_stacks: [ 'cflinuxfs2' ]
 v3_stacks:
```
arjun024, line 390: Also in the config, wouldn't we need some mechanism to filter out version-lines like dotnet-{sdk,runtime,asp} 3.1.X that would never build successfully on cflinuxfs4? (See Ryan's comment in the doc)

pipelines/config/dependency-builds-cflinuxfs4-test.yml
brayanhenao: I thought about leaving the pipeline for a few weeks to do a quick build of the dependencies without interfering with the main pipeline.

The way it is used is basically modifying the listed dependencies when starting work on a new buildpack, i.e. if you need to work with the Nodejs buildpack, the pipeline will only contain the dependencies of this one. These changes will not need to be committed, as this will be the only change (initial structure using .NetCore Buildpack as an example).

This test pipeline would exist until all dependencies are built for cflinuxfs4. After that, it can be deleted from Git and from Concourse.

pipelines/config/dependency-builds.yml
brayanhenao: I was not aware that this change is no longer used. We can eliminate this (I'll delete the references to bionic too).

pipelines/config/dependency-builds.yml
brayanhenao: Totally miss those two dependencies, added.

pipelines/config/dependency-builds.yml
brayanhenao: Yeah, I just updated the PR with the proposed changes to skip building specific version lines in cflinuxfs4.

arjun024 (approved):

PR #159, opened 2022-07-25
Remove pinned version of setuptools in pip installation.

robdimsdale:
In #147 we unpinned the version of `setuptools` in some places but not in all. This PR unpins the remaining versions of `setuptools`, as currently the `pip` build fails due to inconsistent versions of `setuptools`.


brayanhenao:
@robdimsdale I unpinned the version because it was causing some problems with `pipenv` build. Let me double check. If so, then we need to split that method and install the specific version for `pipenv` and `pip`

brayanhenao (approved):

PR #165, opened 2022-08-06
Filter cflinuxfs4 dependencies in buildpacks that does support it yet

brayanhenao:
# Context
In this [PR](https://github.com/cloudfoundry/nodejs-buildpack/pull/474/files#diff-24c51d6f58480f0d1e1b3f4068768cbdea5c9896edb96ca0165f8122224217a8R91) for NodeJS BP, the `node` dependency was updated using the `cflinuxfs4` stack. This stack right now isn't flagged as `cflinuxfs4 available`. 

# Solution
The logic in the task `tasks/update-buildpack-dependency/run.rb` was updated to filter buildpacks that are not available yet for cflinuxfs4.


brayanhenao:
c.c @arjun024

PR #181, opened 2022-10-18
Update Python and Go Buildpack to only keep 1 of each patch version (latest)

brayanhenao:
Go Buildpack and Python Buildpack exceeded the maximum size allowed by CF for uploading a Buildpack (1GB) so I decided to apply the same tactic used in Dotnet Core Buildpack, removing versions of the dependencies to have only the latest version (latest patch).


brayanhenao:
C.C @arjun024 @robdimsdale @KieranJeffreySmart

PR #187, opened 2022-11-02
Adds Node.js v19 version line

ryanmoran:


ryanmoran:
We've decided that it doesn't make sense to support this version line.

PR #197, opened 2022-12-05
Deprecate PHP 7.4.x

brayanhenao:
# PHP 7 Modules shared

https://www.diffchecker.com/0BqLDGON


tasks/check-for-latest-php-module-versions/run.rb
```diff
@@ -57,9 +53,6 @@
 If you're updating cassandra modules (including datastax/cpp-driver) please do so in individual commits, then rebuild appropriate php versions, so integration tests can run in CI with only cassandra changes.
 This will help isolate the php cassandra module change(s) if the changes cause problems.
 
-If the release is compatible with all versions update the php[7|8]-base-extensions.yml file. Otherwise,
```
arjun024, line 60: nitpick: this is still correct right?

arjun024:
Quick search for php7 and I saw this https://github.com/cloudfoundry/buildpacks-ci/blob/master/spec/tasks/build-binary-new/builder_spec.rb#L111. We don't have to necessarily do this now but wondering if you're able to run the spec tests?

tasks/check-for-latest-php-module-versions/run.rb
brayanhenao: In this case, since we don't have two base extension versions we don't need this line.

arjun024: I believe the line is intended to communicate "If the module release is compatible with all of 8.0, 8.1...8.N, simply update the php8-base.yml. Otherwise make updates in php81-patch.yml...php8N-patch.yml". No strong opinions either way

arjun024 (approved):

PR #205, opened 2022-12-16
Add PHP 8.2.x

brayanhenao:


arjun024 (changes requested):
@brayanhenao
- I believe we should add the case for 8.2 here https://github.com/cloudfoundry/buildpacks-ci/blob/7d04c8af18cb270d2ae639e81a6b394da365dacb/tasks/update-buildpack-dependency/run.rb#L200-L212
- What about extension patch files? Do we first build the dependency and then add them later if extensions don't seem to work? I can't recall how we do these things

arjun024 (approved):

PR #208, opened 2023-01-05
Update compile ruby version from 2.5 to 2.7 for bundler dep

brayanhenao:
Bundler versions >= `2.4.x` requires Ruby >= `2.6.0` to be compiled.

### DIFF
--- a/tasks/build-binary-new/builder.rb
+++ b/tasks/build-binary-new/builder.rb
@@ -19,6 +19,13 @@ end
 module DependencyBuild
   class << self
 
+    def setup_ruby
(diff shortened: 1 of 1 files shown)


arjun024:
@brayanhenao I re-triggered the bundler build job after merge and it seems to have [passed](https://buildpacks.ci.cf-app.com/teams/main/pipelines/dependency-builds/jobs/build-bundler-2.x.x/builds/80). But, I see builds to node, php, ruby etc. failing with similar errors. Do you think we should either globally install 2.7 in the task, or even move that logic to binary-builder?

PR #211, opened 2023-01-12
Build Nginx dependencies for cflinuxfs4

brayanhenao:
# Notes

- `openresty` has so many lines right now and we should support only the latest 2. This applies to cflinuxfs4 since deleting 3 version lines from cflinuxfs3 could impact some customers.
- `nginx-static` is added to `builder.rb` even though is not a direct dependency of the Nginx Buildpack but it shares the build logic with `nginx`

### DIFF
--- a/pipelines/config/dependency-builds.yml
+++ b/pipelines/config/dependency-builds.yml
@@ -179,6 +179,7 @@ dependencies:
           - line: 1.19.X
           - line: 1.21.X
     versions_to_keep: 2
+    skip_lines_cflinuxfs4: [ '1.13.X' ]
   nginx-static:
     buildpacks:
       php:
--- a/pipelines/config/dependency-builds.yml
+++ b/pipelines/config/dependency-builds.yml
@@ -173,13 +173,9 @@ dependencies:
     buildpacks:
       nginx:
         lines:
-          - line: 1.13.X
-          - line: 1.15.X
-          - line: 1.17.X
           - line: 1.19.X
           - line: 1.21.X
     versions_to_keep: 2
-    skip_lines_cflinuxfs4: [ '1.13.X' ]
   nginx-static:
     buildpacks:
       php:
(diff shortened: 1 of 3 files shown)


pipelines/config/dependency-builds.yml
```diff
@@ -179,6 +179,7 @@ dependencies:
           - line: 1.19.X
           - line: 1.21.X
     versions_to_keep: 2
+    skip_lines_cflinuxfs4: [ '1.13.X' ]
   nginx-static:
     buildpacks:
       php:
```
arjun024, line 185: Shouldn't there be additions to cflinuxfs4_build_dependencies, cflinuxfs4_dependencies and cflinuxfs4_buildpacks variables?

tasks/build-binary-new-cflinuxfs4/builder.rb
```diff
@@ -532,6 +532,123 @@ def build_yarn
...
+
+    nginx_options = [
+      '--with-cc-opt=-fPIC -pie',
+      '--with-ld-opt=-fPIC -pie -z now',
+      '--with-compat',
+      '--with-stream=dynamic',
+      '--with-http_sub_module',
+    ]
+
+    build_nginx_helper(nginx_options)
```
arjun024, line 545: Do you think it's better to avoid following the pattern build_X for this helper function, as builder.rb already sets a convention that `build_` is a prefix for first class dependency building functions?

arjun024 (changes requested):

pipelines/config/dependency-builds.yml
brayanhenao: Good catch, I'll update the PR

tasks/build-binary-new-cflinuxfs4/builder.rb
brayanhenao: At the time of having to do the function, I thought of the same thing. I wanted to place it outside, maybe in a helper but semantically it should be in Build since that is what the method is doing. So if you have any suggestions on where we could move that to follow the pattern.

tasks/build-binary-new-cflinuxfs4/builder.rb
```diff
@@ -127,6 +127,41 @@ def strip_top_level_directory_from_zip(filename, destination)
...
 
+module DependencyBuildHelper
+  class << self
+    def build_nginx_helper(source_input, custom_options, static=false)
+      artifacts = "#{Dir.pwd}/artifacts"
+      destdir = Dir.mktmpdir
+      Dir.mktmpdir do |dir|
+        Dir.chdir(dir) do
+          Runner.run('wget', source_input.url)
+          # TODO validate pgp
```
TisVictress, line 138: You forgot to implement this validation.

tasks/build-binary-new-cflinuxfs4/builder.rb
```diff
@@ -532,6 +567,92 @@ def build_yarn
...
+
+  def build_openresty
+    artifacts = "#{Dir.pwd}/artifacts"
+    destdir = Dir.mktmpdir
+
+
+    Dir.mktmpdir do |dir|
+      Dir.chdir(dir) do
+        Runner.run('wget', @source_input.url)
+        # TODO validate pgp
```
TisVictress, line 612: Same as above

TisVictress (commented):
LGTM but I think its important to do validation

tasks/build-binary-new-cflinuxfs4/builder.rb
brayanhenao: This was part of the code I took from older `builder.rb`. I don't know what `pgp` means, do you have any context on this?

Ref: https://github.com/cloudfoundry/buildpacks-ci/blob/master/tasks/build-binary-new/builder.rb#L326

arjun024 (approved):

tasks/build-binary-new-cflinuxfs4/builder.rb
brayanhenao: @TisVictress After some quick investigation I found what is PGP an how we can validate the file integrity with it. I update the PR with the changes

PR #219, opened 2023-01-25
rename the test pipeline for cflinuxfs4 to avoid any confusion

arjun024:
This is a pipeline engineers can use to test dependencies and does not build the dependencies that go into the buildpack.

Also remove variables not used

### DIFF
--- a/pipelines/config/test-bed-dependency-builds-fs4.yml
+++ b/pipelines/config/test-bed-dependency-builds-fs4.yml
@@ -13,35 +13,9 @@ dependencies:
       - shiny
       - plumber
     versions_to_keep: 2
+    skip_lines_cflinuxfs4: ['3.6.X']
 
-cflinuxfs4_dependencies: [ 'r']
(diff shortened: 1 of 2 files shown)


arjun024:
@brayanhenao @sophiewigmore @robdimsdale

This is just a suggestion as the pipeline name `dependency-builds-cflinuxfs4` could confuse people to believe that production dependencies are built in this separate pipeline. Any thoughts?

If merged, after-merge actions:
[ ] Fly the pipeline and test with one of the already passed dependency
[ ] Destroy old pipeline
[ ] Update the wiki

robdimsdale (approved):
Yup, I found it confusing when i first looked at it. I think this name is better.

arjun024:
Running the new pipeline here to make sure it works the same: https://buildpacks.ci.cf-app.com/teams/main/pipelines/test-bed-dependency-builds-fs4

brayanhenao (approved):

arjun024:
All post merge tasks done

PR #232, opened 2023-02-08
Dockerfile: remove apt.starkandwayne source as it is no longer available

sophiewigmore:
Signed-off-by: Brayan Henao <bhenao@vmware.com>


brayanhenao:
Same as #231

PR #245, opened 2023-03-07
Update detect-and-upload task for php buildpack packaging

brayanhenao:
# Context

On https://github.com/cloudfoundry/php-buildpack/pull/815 we updated the PHP Buildpack Gemfile required Ruby version to 3.0. Now that the `detect-and-upload` task runs, it fails because of the ruby version used in the `cfbuildpacks/ci` image on which the task is being run.

We cannot update just the image to use a jammy-based one because then it will fail due to the version of the Gemfile/dependencies in this repo [Gemfile](https://github.com/cloudfoundry/buildpacks-ci/blob/master/Gemfile).

# Solution

This is a temporary solution to get the task pass and package the Buildpack. I download the old Ruby2.7 gemfile/gemfile.lock and use that one to package the PHP Buildpack. This only applies to PHP since is the only Buildpack that has the `compile-extensions` folder.

This should be updated when bionic stack is deprecated and this repo Gemfile is updated.

### DIFF
--- a/tasks/detect-and-upload/buildpack-tagger.rb
+++ b/tasks/detect-and-upload/buildpack-tagger.rb
@@ -10,7 +10,6 @@ class BuildpackTagger
     @git_repo_org = git_repo_org
   end
 
-
   def run!
     Dir.chdir(buildpack_dir) do
       tag_to_add = "v#{File.read('VERSION')}".strip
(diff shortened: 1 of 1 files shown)


sophiewigmore:
@brayanhenao can you link me to the failed CI job?

sophiewigmore:
I'm also a little confused about how this task interacts with both the buildpack Gemfile and the Gemfile in this repository. Can you explain a bit more?

sophiewigmore:
The alternative to this PR would be creating an entirely new copy of the buildpacks-ci image based on Jammy, pushign it and using JUST the PHP pipeline's detect and upload task on that image. I think that introduces more tech debt so I'm ok with this change. 

Let's just make sure to file an issue describing the tech debt we've introduced here, and how/when we can remove it.

sophiewigmore (approved):

PR #261, opened 2023-04-19
Remove buildpacks extensions pipeline and unused slack failure hook.

robdimsdale:
This PR removes the `buildpacks-extensions` pipeline and the associated tasks.

* We no longer maintain [stack auditor](https://github.com/cloudfoundry/stack-auditor), and the team that does has their own CI pipeline
* We have deprecated [buildpack-packager](https://github.com/cloudfoundry/buildpack-packager). If we ever make more changes we can refer to this PR for how to create another release.

This PR also removes the broken slack failure hooks, which haven't worked in some time.


arjun024:
@brayanhenao @sophiewigmore Does PHP buildpack use the old packager? you won't need the pipeline right?
https://github.com/cloudfoundry/buildpacks-ci/blob/bd683f63cb46cde124a5839a85d3bf23a50739d5/tasks/detect-and-upload/buildpack-tagger.rb#L55

robdimsdale:
@arjun024 Yes, Php does use the old packager, but we don't intend on making any updates to the old packager and hence no new releases. The job that updates the packager in the php buildpack is paused.

arjun024 (approved):

PR #264, opened 2023-04-20
Update gcc version used to build Nodejs

brayanhenao:
# Context

The latest version of Node `18.X` replaces the URL parser with `ADA` ([Changelog ref](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#replace-url-parser-with-ada)).

The compilation started to fail only on the `cflinuxfs3` stack because ADA requires some `C/C++` headers that are not present in the default `gcc/g++` version provided by Ubuntu.

```
$ gcc -v
gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)
```

# Solution

This required header is present on `gcc/g++  >= 8` and the Nodejs page indicates that every build should have `gcc >= 10.1` ([notes](https://github.com/nodejs/node/blob/main/BUILDING.md#unix-prerequisiteshttps://github.com/nodejs/node/blob/main/BUILDING.md#unix-prerequisites)).

After some experimentation, I found that every build done with `gcc >= 8` fails for errors with the base image when running tests ([ref](https://github.com/cloudfoundry/nodejs-buildpack/actions/runs/4739748884/jobs/8414885768#step:5:125)) and it works with 8, so now that version is installed before the build task starts.

```
$ gcc -v
gcc version 8.4.0 (Ubuntu 8.4.0-1ubuntu1~18.04)
```

# Notes

Also, there is some reformat code done in the file.

### DIFF
--- a/tasks/build-binary-new/builder.rb
+++ b/tasks/build-binary-new/builder.rb
@@ -36,6 +36,15 @@ module DependencyBuild
       Runner.run('rm', '-f', 'get-pip.py')
     end
 
+    def setup_gcc11
+      Runner.run('apt', 'update')
+      Runner.run('apt', 'install', '-y', 'software-properties-common')
+      Runner.run('add-apt-repository', '-y', 'ppa:ubuntu-toolchain-r/test')
+      Runner.run('apt', 'update')
+      Runner.run('apt', 'install', '-y', 'gcc-8', 'g++-8')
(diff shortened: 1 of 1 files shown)


robdimsdale (approved):

robdimsdale:
Leaving this here for node maintainers to take a look at it too cc @cloudfoundry/wg-app-runtime-interfaces-buildpacks-node-js-approvers

PR #268, opened 2023-04-24
Remove unused pipelines.

robdimsdale:
We no longer use this pipeline as we have added support for cflinuxfs4 to all buildpacks.


sophiewigmore:
@brayanhenao do you know if we are OK to remove these?

PR #278, opened 2023-05-05
new dep version lines: tracker story -> github event

arjun024:
When depwatcher detects new version lines, it used to create a story in the tracker project to remind engineers to include it in the dependency-build config. Instead of API call to tracker to create a story, call Github API to send a dispatch of type `new-version-line` to the "cloudfoundry/buildpacks-github-config" repo. This change also requires a dispatch handler at that repo which uses the data to create an issue and add it to the right project. (https://github.com/cloudfoundry/buildpacks-github-config/pull/24)

### DIFF
--- a/pipelines/dependency-builds.yml.erb
+++ b/pipelines/dependency-builds.yml.erb
@@ -108,12 +108,6 @@ end
 def get_git_repo(buildpack)
   bp_uri = "cloudfoundry/#{buildpack}-buildpack"
 end
-
-def read_tracker_story_id(tracker_story_dir)
-  Dir.glob("#{tracker_story_dir}/id").first do |file|
-    return File.read(file).strip
-  end
-end
 %>
 
 resource_types:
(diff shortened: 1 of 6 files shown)


arjun024:
Not fly-ed yet. Needs task merged to fly

pipelines/dependency-builds.yml.erb
```diff
@@ -292,13 +283,11 @@ version_lines = dep['buildpacks'].values.reduce([]) {|sum, bp| sum | get_version
           - name: all-monitored-deps
   <% end %>
   <% if dep_name.downcase == 'node' && line.downcase == 'node-lts' %>
-    - task: create-tracker-story
+    - task: create-new-version-line-story-node
```
sophiewigmore, line 286: why Node specifically?

sophiewigmore: just trying it there first?

arjun024: Not trying out. This is a special task exclusively for node similar to the task on line 35 for all other dependencies.
In my understanding of the file's history, Node is considered special because its LTS dependency alone is needed in Ruby and .NET buildpacks and is accomplished by maintaining a special dependency named `node-lts` in depwatcher. Looks like @brayanhenao implemented this in #130.
In this PR, I haven't looked deeply into rearchitecting the pipeline and just tried to replace tracker with github.

robdimsdale (approved):
LGTM.

Longer-term I'd like to explore if we can use the `gh` CLI to create the request, instead of a ruby script, for a few reasons:
* I think bash wrapping `gh` will be more maintainable for this team
* It's quicker and easier to define and build an image with `gh` binary installed vs a full ruby installation
* The `gh` CLI will handle things like authentication, URI construction, and headers (e.g.`Content-Type`)

However, for now I think we should make the smallest incremental change, so let's merge this.

PR #282, opened 2023-05-15
minor fix in cf-release pipeline

arjun024:
- it looks like git ssh resource won't run without a private_key (partially revert #277)
- Since we haven't yet migrated core-deps team pipelines to main team yet (#271), temporarily copy-in update-pipeline script from the old core-deps repo for current daytoday use.


arjun024:
Already fly-ed

PR #289, opened 2023-06-15
keep updating python patches in node BP

arjun024:
I noticed that the python dep in nodejs buildpack never got updated since it was manually added.[1]

We should probably keep up with patches to aviod CVEs in the shipped buildpack

1: https://github.com/cloudfoundry/nodejs-buildpack/blame/v1.8.11/manifest.yml#L120-L135


arjun024:
I haven't flown it yet

brayanhenao (approved):

PR #295, opened 2023-07-14
depwatcher/miniconda: adapt to new format

arjun024:
- Miniconda changed their versioning from SemVer to CalVer (https://github.com/conda-incubator/ceps/blob/main/cep-8.md)
- Their version is `Year.Month.Patch`, but their upstream artifact is url-versioned `Year.Month.Path-BuildNumber` See example build script: https://github.com/conda/conda-docs/blob/e13a319bd24d9ce754d1f2e9091c7fc522922733/docs/source/create_miniconda_rst.py#L30
- They publish sha256 and not md5 like they used to.
- Publish sha from the upstream page and do not calculate (that's not depwatcher's job) - depwatcher resource simply passes upstream data down.

Fixes https://github.com/cloudfoundry/python-buildpack/issues/761


arjun024:
@cloudfoundry/wg-app-runtime-interfaces-buildpacks-python-approvers

robdimsdale (approved):

PR #305, opened 2023-08-31
fix redeploy task

arjun024:
2 changes:
1) Fix pipeline error on upload release by removing the --fix arg. This seems to come from a bosh error which has been fixed in bosh v277.3.0 (cloudfoundry/bosh/pull/2428) but our environments are still on v276.1.1.
2) Newer cf-deployment versions have removed support for cflinuxfs3 and default deployment simply installs cflinuxfs4. Therefore, flip the case from a pipeline perspective - jobs that need fs3 explicitly asks for it to be added, instead of them asking to add fs4.


arjun024:
Not fly-ed yet due to an lpass issue with my client, but binary-buildpack passes tests with this branch

robdimsdale (approved):

arjun024:
All fly-ed now

PR #306, opened 2023-08-31
cats: defer running readiness healthcheck test

arjun024:
cats 15.1.0 introduced the readiness healthcheck test[1] and is failing in the pipeline. This is because it needs capi, diego etc. from cf-d 32.2.0 but toolsmiths is a few versions behind and we checkout toolsmiths's env context[2] before our deploy.

I'm also not convinced that the buildpacks team needs to run these tests!

--

1. https://github.com/cloudfoundry/cf-acceptance-tests/releases/tag/v15.1.0
2. https://github.com/cloudfoundry/buildpacks-ci/blob/7b677a2baaffc1c2ae4b01504576f3ea9ffc2751/pipelines/cf-release/cf-release.yml#L207

### DIFF
--- a/tasks/cf-release/create-cats-integration-config/run
+++ b/tasks/cf-release/create-cats-integration-config/run
@@ -53,6 +53,7 @@ cat >cats-integration-config/integration-config.json <<-EOF
   "include_v3": false,
   "include_volume_services": false,
   "include_zipkin": false,
+  "readiness_health_checks_enabled": false,
   "stacks": ["cflinuxfs4"]
 }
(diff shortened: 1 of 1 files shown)


arjun024:
If you have reasons to believe that we need these tests, I can add an issue and assign it to myself to revert this change when toolsmith envs support these tests

arjun024:
also ping @brayanhenao - I just happened to see you pinned the resource and ran the pipeline

brayanhenao (approved):

PR #311, opened 2023-09-22
Add setuptools/pip upgrade step when doing the pip setup

brayanhenao:
# Context

The `pipenv` build process started to fail for the `cflinuxfs4` stack. After some investigation, I found that there are some issues with the default version of `setuptools` that comes by default when you install `python3` and `python3-pip`, which is causing problems with the download of dependencies.

Build failure: [132](https://buildpacks.ci.cf-app.com/teams/main/pipelines/dependency-builds/jobs/build-pipenv-latest/builds/132)

# Solution

I have added an upgrade command of `setuptools` to the `setup_python_and_pip` function used to build `pipenv`. This will install the latest available version.

Build succeded with the proposed changes: [133](https://buildpacks.ci.cf-app.com/teams/main/pipelines/dependency-builds/jobs/build-pipenv-latest/builds/133)

### DIFF
--- a/tasks/build-binary-new-cflinuxfs4/builder.rb
+++ b/tasks/build-binary-new-cflinuxfs4/builder.rb
@@ -822,6 +822,7 @@ class DependencyBuild
     def self.setup_python_and_pip
       Runner.run('apt', 'update')
       Runner.run('apt', 'install', '-y', 'python3', 'python3-pip')
+      Runner.run('pip3', 'install', '--upgrade', 'pip', 'setuptools')
     end
 
     def self.prune_dotnet_files(source_input, files_to_exclude, write_runtime = false)


robdimsdale:
Thanks for digging into this @brayanhenao. Do you happen to know how this will interact with [this line](https://github.com/cloudfoundry/buildpacks-ci/blob/1c4143e5d7b2a5d77af7878de03cb798e7b79a72/tasks/build-binary-new-cflinuxfs4/builder.rb#L557) where we download (but don't install) a specific version of `setuptools`?

brayanhenao:
@robdimsdale That's a good question. I think we should delete that line since we already have the dependency downloaded.

robdimsdale:
@brayanhenao we can try that, but we should be prepared for it to fail (in CATS), because sometimes there a tight coupling between the version of setuptools and the version of pip. That's why we pin to a specific version anyway.

PR #312, opened 2023-09-25
dedicated brats: run on latest LTS version that supports fs4

arjun024:
The version used at moment for dedicated brats pipeline (2.11) do not support cflinuxfs4, and therefore cannot run fs4 brats tests. Update env to 4.0 in accordance with the intention of the pipeline.

The lts-brats tests of the buildpacks pipeline (separate from this) currently runs the most popular flavor[1] 2.11 on cflinuxfs3. When that goes out of support, we will update it to 4.0 as well.

There's also another brats test job in the buildpacks pipeline that runs both cflinuxfs3&4 tests on a cf-d env.

1. [private slack link](https://vmware.slack.com/archives/CU6RWNDRR/p1695416295328009?thread_ts=1695407962.392299&cid=CU6RWNDRR)

### DIFF
--- a/pipelines/brats.yml.erb
+++ b/pipelines/brats.yml.erb
@@ -11,11 +11,11 @@
 "ruby" => %w(cflinuxfs4),
 "staticfile" => %w(cflinuxfs4),
 } %>
-<% pas_versions = {
-  'latest' => '2_13',
-  'n-1' => '2_12',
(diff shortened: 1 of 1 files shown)


arjun024:
flyed

robdimsdale (approved):

PR #315, opened 2023-10-03
ci images: update bosh-cli to assume AWS role

arjun024:
See https://bosh.io/docs/s3-release-blobstore/#assume-role


arjun024:
Both updated in dockerhub (latest)

PR #319, opened 2023-10-05
Update php modules october (Little refactor to bump script included)

brayanhenao:
Besides the normal `.yml` files updates, I decided to update/refactor a little bit the bumping scripts (before it did not support certain dependencies so when a bumping occurs, you have to search manually by visiting the URL to see if there were any bumps).

As you can see in the code, it parses the `HTML` of the page that lists the releases (I could not find any endpoint that gave me the releases in another format) so it may be prone to errors due to changes you may make to the pages, but normally it will not require a major change. For the moment, leaving it like this saves a lot of time by not having to manually visit and search for these dependencies.


brayanhenao:
CC @robdimsdale @sophiewigmore

PR #321, opened 2023-10-23
s3 buildpacks-team user: move to access via cloudgate

arjun024:
This is the user that owns the s3 bucket "pivotal-buildpacks" used for both dependency storage, and online buildpack bosh release blob storage.

See [issue](https://github.com/pivotal-cf/tanzu-buildpacks/issues/292) for more details.


arjun024:
Fly-ed the changes

robdimsdale (approved):

PR #322, opened 2023-10-25
fix semver resource's assume role param

arjun024:
see https://github.com/concourse/semver-resource#s3-driver


arjun024:
Fly-ed

robdimsdale (approved):

PR #326, opened 2023-10-27
Update cflinuxfs3-dev dockefiles with the ESM packages to be removed

brayanhenao:
In #324, I initially made changes to include the removal of some additional ESM packages. However, after thorough debugging, it became evident that only `libsnmp-dev` is necessary to resolve the PHP Compilation errors.


brayanhenao:
C.C @robdimsdale

PR #327, opened 2023-10-27
Fix error in PHP Update dependency logic (related to cflinuxfs3-dev)

brayanhenao:
When we introduced the `cflinuxfs3-dev` stack in #300, the `tasks/update-buildpack-dependency` task began encountering failures, despite the updates introduced in #303. To address this issue, I made a change to ignore the `cflinuxfs3.json` file in PHP binary-builds, ensuring that older versions are not selected over those specified in `cflinuxfs3-dev.json`.


brayanhenao:
C.C @arjun024

PR #339, opened 2024-01-02
cf-target: do not assume api url from env name

arjun024:
Looks like the api url is now named `api.$ENV_NAME.$RANDOM_STRING.shepherd.lease`


arjun024:
@cloudfoundry/wg-app-runtime-interfaces-buildpacks-tooling-approvers @brayanhenao Feel free to review and revert

Issue #309, opened 2023-09-07
Automatically build with the latest versions of Go


robdimsdale:
In #308 we manually bumped the version of Go used in the Dockerfiles - and hence used to build the buildpacks - to the latest version.

We did this because we were notified that the buildpacks were built with a version of Go with known CVEs. If we hadn't been notified, we wouldn't have made the change. Once we were notified of the CVEs, it still took a while to track down where Go is installed.

We should setup some automation to automatically use a new version of Go when one is released. That way we don't have to worry about CVEs, and we don't have to spend time and cognitive effort tracking down where Go is installed.

This could look like automating a PR like #308, or we could explore alternative solutions that don't rely on a pre-installed Go in a docker image (e.g. if we use github actions we could use the action to install the latest stable version of Go).

arjun024:
**Suggestion of a possible implementation**
In dockerfile:
```
RUN export GO_VERSION=$(wget -qO- https://golang.org/dl/?mode=json | grep -oP '"version": "\K([^"]+)' | head -n 1) && \
    wget -q https://golang.org/dl/go$GO_VERSION.linux-amd64.tar.gz && \
    tar -C /usr/local -xzf go$GO_VERSION.linux-amd64.tar.gz
```

and the docker-build workflow to check `.[0].version` https://golang.org/dl/?mode=json and compare against a version stored locally in the repo

brayanhenao:
Implemented the solution in #355

PR #353, opened 2024-02-20
ci images: Bump go to 1.22.0

arjun024:


sophiewigmore:
Good call on this PR, but do you think we should generalize to just use the latest to knock out https://github.com/cloudfoundry/buildpacks-ci/issues/309 ?

arjun024:
@sophiewigmore Good idea. I have an implementation suggestion [here](https://github.com/cloudfoundry/buildpacks-ci/issues/309#issuecomment-1955112007)

PR #367, opened 2024-04-25
Use our own shepherd-pool instead of the public one

brayanhenao:
Screenshot 2024-04-29 at 10 25 59 AM


brayanhenao:
Changes already flyed.

PR #385, opened 2024-06-28
Remove buildpacks-site pipelines

robdimsdale:
- it has not been updated since December 2021, and we do not even know where this site is hosted.


pipelines/templates/buildpack.yml.erb
```diff
@@ -101,7 +101,7 @@ resources: #####################################################################
   - name: cf-deployment
     type: git
     source:
-      uri: https://github.com/cloudfoundry/cf-deployment
```
arjun024, line 104: typo

arjun024 (changes requested):

pipelines/templates/buildpack.yml.erb
robdimsdale: wow that's a bonafide typo!

robdimsdale: clearly i was trying to run the `gh` command in my terminal 😅

arjun024 (approved):

PR #403, opened 2024-07-15
Switch to using buildpacks.cloudfoundry.org bucket

robdimsdale:
- this is necessary for the CFF to take over the CDN that sits in front of this namespace.
- there is a lot of specific nuance, but the summary is that the CDN must now point to a bucket that has the same name
- this does not affect consumers because they were already downloading dependencies from the CDN under the buildpacks.cloudfoundry.org host

Changes have already been fly'd

### DIFF
--- a/public-config.yml
+++ b/public-config.yml
@@ -6,7 +6,6 @@ buildpack-packager-git-uri-public: https://github.com/cloudfoundry/buildpack-pac
 buildpack-packager-git-uri: git@github.com:cloudfoundry/buildpack-packager.git
 buildpack-release-candidates-bucket: buildpack-release-candidates
 buildpackapplifecycle-git-branch: develop
-buildpacks-binaries-s3-bucket: pivotal-buildpacks
 buildpacks-ci-git-uri-public-branch: master
 buildpacks-ci-git-uri-public-develop-branch: develop
 buildpacks-ci-git-uri-public: https://github.com/cloudfoundry/buildpacks-ci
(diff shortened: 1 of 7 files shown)


tasks/finalize-buildpack/task.yml
```diff
@@ -7,14 +7,14 @@ image_resource:
 inputs:
   - name: buildpacks-ci
   - name: buildpack
-  - name: pivotal-buildpacks-stack0
-  - name: pivotal-buildpacks-stack1
+  - name: buildpacks.cloudfoundry.org-stack0
```
robdimsdale, line 10: It is not clear to me if the `pivotal-buildpacks-stackX` naming is important or not, and therefore I can't tell if this will break the `finalize-buildpack` task. Given there's no easy way to test this, let's just keep an eye out for this task breaking when it is invoked.

public-config.yml
```diff
@@ -6,7 +6,6 @@ buildpack-packager-git-uri-public: https://github.com/cloudfoundry/buildpack-pac
 buildpack-packager-git-uri: git@github.com:cloudfoundry/buildpack-packager.git
 buildpack-release-candidates-bucket: buildpack-release-candidates
 buildpackapplifecycle-git-branch: develop
-buildpacks-binaries-s3-bucket: pivotal-buildpacks
```
robdimsdale, line 9: This is removed because it was only used in one place (and hence inlined in that location), and I think the indirection is not helpful.

arjun024 (approved):

PR #406, opened 2024-07-16
Merge brats job into corresponding integration test job.

robdimsdale:
- this saves on the number of environments we need to claim at the expense of adding a few more minutes to the duration of the job.
- Claiming environments can take a long time, and releasing them can be flaky. So this change increases the predictability of the CI, by reducing the number of times we might have to wait a long time for an environment, and the number of opportunity for lease release to fail.

Changes have been fly'd to ~~**most but not all**~~ all buildpack pipelines. I didn't fly them all because some were running and I didn't want to cut them off before they finished.


arjun024 (approved):

pipelines/templates/buildpack.yml.erb
```diff
@@ -522,15 +453,34 @@ jobs: ##########################################################################
...
+            <% if language == 'php' %>
+            COMPOSER_GITHUB_OAUTH_TOKEN: '((composer-github-oauth-token))'
+            <% end %>
+            <% if buildpacks[language]['skip_docker_start'] %>
+            SKIP_DOCKER_START: true
+            <% else %>
+          privileged: true
+            <% end %>
+        ensure:
+          task: delete-cf-space
```
arjun024, line 482: It looks like we need the the create-cf-space task so that the input dir to the task exists. https://buildpacks.ci.cf-app.com/teams/main/pipelines/go-buildpack/jobs/specs-edge-integration-develop-cflinuxfs3/builds/108

robdimsdale:
All pipelines have been fly'd and I've seen this change go green on multiple pipelines. Merging.

robdimsdale:
Dang. My change overrode the approval @arjun024 - could you take another look please.

arjun024 (approved):

PR #407, opened 2024-07-18
remove build/test on windows2016 stack

arjun024:
Cf-d removed support for stack[1]. windows2016 is just an alias for windows stack kept for legacy reasons. The CF docs already state[2] windows2016 as an unsupported stack, and I believe we have warned users to switch for years.

1. https://github.com/cloudfoundry/cf-deployment/pull/1156
2. https://docs.cloudfoundry.org/devguide/deploy-apps/windows-stacks.html#-available-stacks

### DIFF
--- a/pipelines/cf-release/cf-release-config.yml
+++ b/pipelines/cf-release/cf-release-config.yml
@@ -4,7 +4,7 @@
 
 supported_languages:
 - name: binary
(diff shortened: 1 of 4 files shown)


arjun024:
All fly-ed with `./bin/update-pipelines`

robdimsdale (approved):

PR #420, opened 2024-11-18
dotnet: rm unsupported 6.x, add 9.x

arjun024:
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core


arjun024:
already flyed

PR #427, opened 2025-03-11
depwatcher/.net: Concourse having trouble with windows.net

arjun024:
so use a mirror on github


arjun024:
already pushed the image to `coredeps/depwatcher`

PR #441, opened 2025-08-27
depwatcher/php: the endpoints have changed

arjun024:
use jq directly. I have commented out the crystal tests for php for now as it didn't seem worth the time to fill out for external calls


arjun024:
already built and pushed to dockerhub. Pipeline is able successfully retrieve versions

PR #448, opened 2025-10-02
Fix cfbuildpacks/ci docker image access task.yml

ivanovac:
Fix the cfbuildpacks/ci docker image access in task.yml


ramonskie (approved):
lgtm

PR #449, opened 2025-10-03
bump to use for ruby 3.4.x

ramonskie:
as i could not install ruby 2.x due to older openssl requirements.

this will make the code compatible with ruby 3.x


ivanovac (approved):
Looks good to me

PR #478, opened 2025-10-27
Automate buildpack version bumping using git-based semver resource

ramonskie:
After PR #473 removed the buildpack-to-master job, dependency update PRs would merge but ship-it would fail because VERSION files weren't being bumped, causing attempts to create artifacts for existing tags.

This adds a semver resource that:
- Uses the buildpack repo's VERSION file as the version state
- Automatically bumps patch version on each master merge
- Updates detect-and-upload to use the bumped version
- Tags releases with the semver version instead of VERSION file
- Enables auto-generated GitHub release notes
- Makes CHANGELOG optional in buildpack finalizer

This enables fully automated dependency releases without manual VERSION file management.


ivanovac (approved):
LGTM

PR #595, opened 2026-03-30
Use version: latest instead of version: every for dependency source resources

ramonskie:
## Problem

Using `version: every` on monitored dependency resources (e.g. `source-rserve-latest`) causes `update-*` jobs to get permanently stuck when an intermediate version of a monitored dep is released but never flows through the corresponding `build-*` job.

**Example**: `rserve` released `1.8.18` between two `build-r` runs. The `build-r` job jumped directly from `1.8.17` to `1.8.19`, so `1.8.18` never had a `build-r` build associated with it. With `version: every`, Concourse requires each version to satisfy the `passed` constraint sequentially — so the `update-r` jobs were permanently blocked even after `build-r` succeeded with `1.8.19`.

## Fix

Change `version: every` to `version: latest` for all dependency source resource `get` steps. This means Concourse only requires the **latest** available version to have passed through the upstream job, rather than every intermediate version.

The `update-*` jobs are not lossy with this change — they still trigger on every new R (or other primary dependency) version via the `source-r-X.X.x` resource. The monitored deps (`rserve`, `forecast`, `shiny`, `plumber`) are used to ensure the built binary is tested with up-to-date packages, so tracking only the latest is sufficient.

### DIFF
--- a/pipelines/dependency-builds/pipeline.yml
+++ b/pipelines/dependency-builds/pipeline.yml
@@ -243,7 +243,7 @@ jobs:
     - get: source
       resource: #@ "source-{}-latest".format(dep_name.lower())
       trigger: true
-      version: every
+      version: latest
   - do:
     - task: create-new-version-line-story
       file: buildpacks-ci/tasks/create-new-version-line-story/create-new-version-line-story.yml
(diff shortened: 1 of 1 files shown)


ivanovac (approved):
lgtm

PR #596, opened 2026-03-30
fix: make any-stack build stack configurable

ramonskie:
## Problem

The `build-binary-any-stack` CI job was failing with:

```
awk: cannot open binary-builder/stacks/any-stack.yaml (No such file or directory)
```

`any-stack` is a build label (meaning "this binary is stack-neutral"), not a real stack. `any-stack.yaml` does not exist and should never be created. The pipeline already runs `any-stack` builds on the `cflinuxfs4` image, but that mapping was:

- Hardcoded in two places in `pipeline.yml`
- **Not propagated to `build.sh` at all** — causing the failure

## Solution

Introduce `any_stack_build_stack` as a single, explicit configuration value. When migrating `any-stack` builds to `cflinuxfs5` in the future, only this one value needs to change.

## Changes

| File | Change |
|------|--------|
| `pipelines/dependency-builds/config.yml` | Add `any_stack_build_stack: cflinuxfs4` — single source of truth |
| `pipelines/dependency-builds/pipeline.yml` | Replace both hardcoded `"cflinuxfs4" if stack == "any-stack"` with `data.values.any_stack_build_stack`; pass `ANY_STACK_BUILD_STACK` param to the build task |
| `tasks/build-binary/build.yml` | Declare `ANY_STACK_BUILD_STACK: cflinuxfs4` param (safe default if task is run outside the pipeline) |
| `tasks/build-binary/build.sh` | When `STACK=any-stack`, resolve `BUILD_STACK` from `ANY_STACK_BUILD_STACK` before reading the stack YAML and invoking `binary-builder`; add comment that `any-stack.yaml` must never be created |

## To migrate to cflinuxfs5 in the future

Change one line in `config.yml`:

```yaml
any_stack_build_stack: cflinuxfs5
```

That's it — `pipeline.yml`, `build.yml`, and `build.sh` all follow automatically.

### DIFF
--- a/pipelines/dependency-builds/config.yml
+++ b/pipelines/dependency-builds/config.yml
@@ -459,6 +459,9 @@ dependencies:
     any_stack: true
     versions_to_keep: 2
 build_stacks: [ 'cflinuxfs4', 'cflinuxfs5' ]
+# Stack used to build any_stack: true dependencies (stack-neutral binaries).
+# When a new stack becomes the default, update this one value.
+any_stack_build_stack: cflinuxfs4
 windows_stacks: [ 'windows' ]
 
 #! only check deprecation dates for dotnet-runtime as they are redundant for sdk and aspnetcore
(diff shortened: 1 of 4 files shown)


tnikolova82 (approved):
lgtm

PR #598, opened 2026-03-31
fix: pass BUILD_STACK to binary-builder so any-stack builds don't fail

ramonskie:
## Problem

After merging #596, `build-binary-any-stack` jobs still failed:

```
binary-builder: loading stack "any-stack": open binary-builder/stacks/any-stack.yaml: no such file or directory
```

`BUILD_STACK` was resolved correctly from `ANY_STACK_BUILD_STACK` inside the go-bootstrap block, but it was **scoped to that block**. The binary-builder invocation below still used `--stack "${STACK}"` (= `any-stack`), causing `binary-builder` to call `stack.Load("any-stack")` and fail.

## Fix

Move `BUILD_STACK` resolution **outside** the `if ! command -v go` block so it is available throughout the script, then pass `--stack "${BUILD_STACK}"` to binary-builder.

## Why the any-stack label is still preserved

The `any-stack` label in artifact filenames does **not** come from `--stack`. It is hardcoded in each recipe's `ArtifactMeta.Stack` field (e.g. `passthrough.go`, `simple.go`). The builds-artifacts JSON filename also still uses `${STACK}` (= `any-stack`). Nothing changes for downstream consumers.

### DIFF
--- a/tasks/build-binary/build.sh
+++ b/tasks/build-binary/build.sh
@@ -4,7 +4,8 @@
 # Responsibilities:
 #   0. Bootstrap Go toolchain (cflinuxfs* images are rootfs images — no Go or Python present).
 #   1. Compile binary-builder from source.
-#   2. Run: binary-builder build --stack $STACK --source-file source/data.json --stacks-dir binary-builder/stacks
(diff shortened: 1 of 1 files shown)


ivanovac (approved):
lgtm

tnikolova82 (approved):
lgtm

PR #600, opened 2026-04-01
Enable cflinuxfs5 stack across all buildpack pipelines

tnikolova82:
## Summary

- Enable cflinuxfs5 stack in all 11 buildpack values files (was commented out with `TODO: re-enable when cf-runtime cflinuxfs5 support is ready`)
- Guard the `version-stack-cflinuxfs5` semver resource in `pipeline.yml` with a ytt conditional so it is only defined for buildpacks that include cflinuxfs5 in their stacks list

## Changes

**Values files** (11 files): Uncomment `- cflinuxfs5` in `buildpack.stacks`:
- `apt-values.yml`, `binary-values.yml`, `dotnet-core-values.yml`, `go-values.yml`, `java-values.yml`, `nginx-values.yml`, `nodejs-values.yml`, `php-values.yml`, `r-values.yml`, `ruby-values.yml`, `staticfile-values.yml`

**`pipeline.yml`**: Wrap `version-stack-cflinuxfs5` resource definition in `#@ if "cflinuxfs5" in data.values.buildpack.stacks:` / `#@ end` to avoid defining an unused resource for buildpacks that don't list cflinuxfs5.

### DIFF
--- a/pipelines/buildpack/apt-values.yml
+++ b/pipelines/buildpack/apt-values.yml
@@ -6,7 +6,7 @@ organization: cloudfoundry
 buildpack:
   stacks:
   - cflinuxfs4
-  #! - cflinuxfs5  #! TODO: re-enable when cf-runtime cflinuxfs5 support is ready
+  - cflinuxfs5
   product_slug: ""
   skip_docker_start: true
   skip_brats: true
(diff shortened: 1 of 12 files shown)


ivanovac (approved):
lgtm

PR #602, opened 2026-04-02
Fix: Add missing version field to build artifacts JSON

ivanovac:
## Summary

This PR fixes a critical bug preventing PR generation for buildpack dependency updates after the filename format fix (#601) was merged.

## Problem

After commit d0299cea fixed the build artifact filename format, builds now create correctly-named files (e.g., `1.29.2.3-cflinuxfs4.json`), but **no PRs are being generated** because the build artifact JSON is missing the required `version` field.

### Root Cause

The `update-buildpack-dependency` script (`run.rb:94-95`) expects each build artifact JSON to contain a `version` field:

```ruby
version = builds[stack]['version']  # Returns nil when field is missing
next unless version                 # Skips processing when nil
```

The Go binary-builder's `build.sh` was not including this field in the jq output (lines 104-110), causing:
1. Build artifacts are created successfully ✅
2. Update job finds the files via glob pattern ✅
3. Script reads the JSON but finds `version = nil` ❌
4. Guard clause `next unless version` skips processing ❌
5. `rebuilt` array remains empty ❌
6. Early-exit at line 139: "SKIP: No build artifacts found" ❌
7. No PR is created ❌

## Solution

Add the `version` field to the build artifact JSON output in `tasks/build-binary/build.sh`:

```bash
jq '{
  version:          .version,    # ← ADDED
  url:              (.url // ""),
  sha256:           (.sha256 // ""),
  source:           (.source // {}),
  source_sha256:    (.source.sha256 // ""),
  sub_dependencies: (.sub_dependencies // {})
}' ...
```

This matches the format produced by the old Ruby builder and expected by `run.rb`.

### Before Fix
- Build artifacts created: ✅
- PRs generated: ❌
- Affected: **ALL** dependencies built after March 30, 2026 (nginx, openresty, ruby, php, pipenv, hwc, tomcat, rubygems, etc.)

### After Fix
- Build artifacts created: ✅
- PRs generated: ✅
- Dependencies can receive cflinuxfs5 (and all stack) updates

## Testing Plan

1. Deploy updated `build.sh` to dependency-builds pipeline
2. Trigger a rebuild for openresty 1.29.2.3
3. Verify build artifact JSON now includes `"version": "1.29.2.3"`
4. Verify update-buildpack-dependency job creates PR with cflinuxfs5 entries
5. Trigger rebuilds for other affected dependencies

## Related

- Depends on: #601 (filename format fix)
- Fixes the issue reported where nginx/openresty builds complete but no PRs are created
[...]

PR #603, opened 2026-04-02
fix: move rsync seeding before artifact write to prevent overwrite

tnikolova82:
## Summary

- Fixes a bug where `rsync -a builds/ builds-artifacts/` overwrites the newly-generated artifact JSON with the stale copy from the `builds` git resource, causing rebuild jobs to silently produce no changes

## Problem

In `build.sh`, the artifact JSON was written to `builds-artifacts/` at step 5 (line 111), but the rsync seeding from the `builds` git resource happened later at step 7 (line 144). This ordering meant:

1. Step 5: Write new artifact JSON (with `version` field, per #602) to `builds-artifacts/binary-builds-new/<dep>/<version>-<stack>.json`
2. Step 7: `rsync -a builds/ builds-artifacts/` copies the **old** file from the `builds` git resource, **overwriting** the new one
3. Step 7: `git diff --cached --quiet` → no changes detected → "No changes to commit"
4. The update job then finds the old artifact (missing `version` field) and prints "SKIP: No build artifacts found for this version"

This was confirmed in today's `build-nginx-1.28.x` build #9 logs:
```
[task] Wrote builds-artifacts to builds-artifacts/binary-builds-new/nginx/1.28.3-cflinuxfs5.json
[task] Seeding builds-artifacts from builds git repo...
[task] No changes to commit in builds-artifacts
```

### Impact

Every rebuild of an existing dependency silently produces a no-op commit, preventing the update jobs from creating PRs with cflinuxfs5 entries. This blocked all dependency updates since the Go builder migration (March 12, 2026).

## Fix

Move the rsync seeding step to **before** writing the artifact JSON, so the new file correctly overwrites the stale copy instead of the other way around:

```
Step 5: rsync -a builds/ builds-artifacts/     ← seed git repo first
Step 6: jq '...' > builds-artifacts/.../X.json ← new file overwrites stale copy
Step 8: git add . && git commit                ← detects the change
```

## Related

- Depends on: #601 (filename format fix) and #602 (missing `version` field fix)
- Completes the fix chain: #601 fixed filenames → #602 added `version` field → this PR fixes the write ordering
- After merge: rebuild jobs will correctly commit updated artifacts, and update jobs will create PRs with cflinuxfs5 entries

## Files Changed

- `tasks/build-binary/build.sh` — Move rsync seeding from after artifact write to before it; renumber steps accordingly

### DIFF
--- a/tasks/build-binary/build.sh
+++ b/tasks/build-binary/build.sh
@@ -8,9 +8,10 @@
 #      ($BUILD_STACK equals $STACK unless STACK=any-stack, in which case ANY_STACK_BUILD_STACK is used)
 #   3. Read the JSON summary from the output file.
 #   4. Move the artifact from CWD to artifacts/.
-#   5. Write builds-artifacts/binary-builds-new/<dep>/<dep>-<version>-<stack>.json
-#   6. Write dep-metadata/<artifact>_metadata.json
-#   7. Optionally commit the builds-artifacts changes to git (when SKIP_COMMIT != true).
+#   5. Seed builds-artifacts from builds git repo (so it is a valid git clone).
+#   6. Write builds-artifacts/binary-builds-new/<dep>/<version>-<stack>.json
+#   7. Write dep-metadata/<artifact>_metadata.json
+#   8. Optionally commit the builds-artifacts changes to git (when SKIP_COMMIT != true).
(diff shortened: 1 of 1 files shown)


ivanovac (approved):
lgtm

PR #610, opened 2026-04-06
fix: compute source SHA256 in rubygems_cli depwatcher

tnikolova82:
## Summary

- Fix the `rubygems_cli` depwatcher to compute source SHA256 hash when fetching version details, matching the pattern used by other watchers (python, jruby, php, nginx, etc.)
- This resolves empty `source_sha256` fields in buildpack manifest entries produced after the Go binary-builder migration

## Root Cause

The `RubygemsCLIWatcher.In()` method was constructing the download URL deterministically but never downloading the source or computing its SHA256 hash. Other depwatchers like `python`, `jruby`, and `php` call `base.GetSHA256()` to download and hash the source.

The old Ruby binary-builder computed source hashes during the build phase, masking this gap. The Go binary-builder rewrite propagates whatever the depwatcher provides but does not independently compute source hashes, exposing the missing SHA256.

## Changes

- `rubygems_cli.go`: `In()` now downloads the source tarball and computes SHA256 using `base.GetSHA256()`
- `rubygems_cli_test.go`: Updated `In` tests to mock the download response and verify SHA256 is computed; added error case test

### DIFF
--- a/dockerfiles/depwatcher-go/pkg/watchers/rubygems_cli.go
+++ b/dockerfiles/depwatcher-go/pkg/watchers/rubygems_cli.go
@@ -60,10 +60,18 @@ func (w *RubygemsCLIWatcher) Check() ([]base.Internal, error) {
 	return versions, nil
 }
 
-// In returns the download URL for a specific RubyGems CLI version.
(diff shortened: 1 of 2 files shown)


ivanovac (approved):
lgtm

PR #615, opened 2026-04-14
Depwatcher fixes

ramonskie:
this will fix the fetch all version for zulu
so `source-zulu-latest` resource will actually fetch the latest versions for zulu

also fixed tests fixtures as they where using the specs from the old crystal based depwatcher


tnikolova82 (approved):
lgtm

PR #616, opened 2026-04-17
[dependency-builds] remove depreacted php 8.1 and added 8.4 and 8.5

ramonskie:
this will remove the depreacted php 8.1 from the dependecy-builds pipeline
and adds 8.4 and 8.5

see: https://www.php.net/supported-versions.php

please not that im not sure which deprecation date we should set.
the end of security updates (which basically means EOL)
or the end of the updates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 private.yml
 .vagrant/
 .rspec
+.opencode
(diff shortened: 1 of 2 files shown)


ivanovac (approved):
lgtm

PR #617, opened 2026-04-19
manage cf-env 3.x.x/4.x.x dependencies

zakerf:
related to https://github.com/cloudfoundry/java-buildpack/pull/1226

### DIFF
--- a/pipelines/dependency-builds/config.yml
+++ b/pipelines/dependency-builds/config.yml
@@ -35,6 +35,18 @@ dependencies:
     source_type: rubygems
     any_stack: true
     versions_to_keep: 2
+  cfenv:
+    buildpacks:
+      java:
+        lines:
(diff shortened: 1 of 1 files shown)


pipelines/dependency-builds/config.yml
```diff
@@ -35,6 +35,18 @@ dependencies:
...
     versions_to_keep: 2
+  cfenv:
+    buildpacks:
+      java:
+        lines:
+          - line: 3.X.X
+          - line: 4.X.X
+    source_type: github_releases
+    source_params:
+      - 'repo: /pivotal-cf/java-cfenv'
```
ramonskie, line 46: should be without starting `/`

zakerf: removed it

ramonskie (approved):

PR #625, opened 2026-04-23
Fix maven SHA512 fallback: use GetRaw to actually inspect status code

ramonskie:
## Problem

PR #624 attempted to fix the maven watcher falling back from SHA512 to SHA1 (Maven Central only provides `.sha1`, not `.sha512`). However the fix did not work in production — the Concourse build continued to fail with:

```
error: fetching version data: failed to fetch SHA512: request failed with status 404: https://repo1.maven.org/maven2/io/pivotal/cfenv/java-cfenv/3.5.1/java-cfenv-3.5.1.jar.sha512
```

## Root Cause

`GetWithHeaders` in `client.go` returns an **error** (not a response) for any non-2xx status code, including 404:

```go
if resp.StatusCode < 200 || resp.StatusCode >= 300 {
    resp.Body.Close()
    return nil, fmt.Errorf("request failed with status %d: %s", resp.StatusCode, url)
}
```

So in `maven.go`, when the SHA512 probe gets a 404, `err != nil` is true and the function returns immediately — the `resp.StatusCode == http.StatusNotFound` fallback check is **never reached**.

## Fix

Add a `GetRaw` method to `HTTPClient` and `HTTPClientImpl` that returns the response regardless of status code (no non-2xx error). `maven.go` now uses `GetRaw` for the SHA512 probe so it can inspect the 404 itself and fall through to the SHA1 fallback.

`GetWithHeaders` is refactored to delegate to `GetRaw` + add the status check on top — no behaviour change for any existing watcher.

## Verification

Tested locally against Maven Central:

```
$ echo '{"source":{"type":"maven","name":"java-cfenv","uri":"https://repo1.maven.org/maven2","group_id":"io.pivotal.cfenv","artifact_id":"java-cfenv"},"version":{"ref":"3.5.1"}}' | /tmp/test-in /tmp/out
{"ref":"3.5.1","url":"https://repo1.maven.org/maven2/io/pivotal/cfenv/java-cfenv/3.5.1/java-cfenv-3.5.1.jar","sha1":"e35e6af21d26b75f70289b2bb31725d345dbb6a7"}
```

All 333 tests pass.

### DIFF
--- a/dockerfiles/depwatcher-go/internal/factory/factory_test.go
+++ b/dockerfiles/depwatcher-go/internal/factory/factory_test.go
@@ -41,6 +41,10 @@ func (m *mockHTTPClient) GetWithHeaders(url string, headers http.Header) (*http.
 	return m.Get(url)
 }
 
+func (m *mockHTTPClient) GetRaw(url string, headers http.Header) (*http.Response, error) {
+	return m.Get(url)
+}
+
 func newMockHTTPClient() *mockHTTPClient {
 	return &mockHTTPClient{
 		responses: make(map[string]string),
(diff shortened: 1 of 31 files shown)


ivanovac (approved):
lgtm

PR #622, opened 2026-04-22
Remove build-binary-new tasks and update stale references

ramonskie:
## Summary

The `build-binary` task is now stack-agnostic, superseding the per-stack `build-binary-new-cflinuxfs{4,5}` tasks and the original `build-binary-new` task.

## Changes

- **Remove** `tasks/build-binary-new`, `tasks/build-binary-new-cflinuxfs4`, and `tasks/build-binary-new-cflinuxfs5` directories (46 files)
- **`README.md`**: Update "Making Changes to Build Scripts" to point to `tasks/build-binary/` and note the task is stack-agnostic; drop the deprecated `cflinuxfs2` `fly` example
- **`pipelines/dependency-builds/README.md`**: Update "How to add a new stack" — no new task directory needed per stack since `build-binary` is stack-agnostic
- **`scripts/php-modules/README.md`** and **`update.rb`**: PHP extension YAMLs moved to `binary-builder/internal/php/assets/`; use `BINARY_BUILDER_DIR` env var (defaults to sibling directory) instead of a hardcoded relative path; add `php84` and `php85` extension files to the processed list

PR #629, opened 2026-05-11
Add client-certificate-mapper dependency for java bp

kiril-keranov:
Add one of the remaining dependencies `client-certificate-mapper` for the java buildpack, more info on still not covered dependencies [here](https://github.com/cloudfoundry/buildpacks-ci/blob/master/docs/missing-java-dependencies-analysis.md). The dependency does not have any specifics with regard to cflinuxfs4/5 , should be added for both.
- It is using [maven](https://github.com/cloudfoundry/buildpacks-ci/blob/master/dockerfiles/depwatcher-go/pkg/watchers/maven.go) dependency watcher, so no new watchers added.
- wondering whether `tasks\build-binary-new-cflinuxfs[4/5]/builder.rb` should also be adjusted? As far as I saw the `java-cfenv` dependency using `maven` watcher is also not added

### DIFF
--- a/pipelines/dependency-builds/config.yml
+++ b/pipelines/dependency-builds/config.yml
@@ -35,6 +35,18 @@ dependencies:
     source_type: rubygems
     any_stack: true
     versions_to_keep: 2
+  client-certificate-mapper:
+    buildpacks:
+      java:
+        lines:
+          - line: 2.X.X
+    source_type: maven
+    source_params:
+      - 'uri: https://repo1.maven.org/maven2'
+      - 'group_id: org.cloudfoundry'
+      - 'artifact_id: client-certificate-mapper'
+    any_stack: true
+    versions_to_keep: 1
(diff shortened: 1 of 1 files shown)


kiril-keranov:
@ramonskie @tnikolova82 @ivanovac can you please check? I haven't updated the `tasks\build-binary-new-cflinuxfs[4/5]/builder.rb` as I saw dependencies using the `maven` watcher like `java-cfenv` are also not mentioned, rather dependencies with their own resolvers are added to the `builder.rb` if my understanding is correct. I've added only this one currently to acquaint with the pipelines

ivanovac (approved):
If this fits your needs and it helps builds the other dependencies, it is fine by me.

ramonskie:
https://github.com/cloudfoundry/buildpacks-ci/pull/622 removes the build-binary-new tasks as we do not use that anymore

PR #638, opened 2026-05-20
Add custom watchers and adjust dependency config

kiril-keranov:
Add more java buildpack dependencies to the dependency-builds pipeline.
- `jrebel` java agent - add to dependency config and custom dependency watcher.  No issues for cflinuxfs5, it bundles a native libjrebel64.so for Linux x86-64, but its maximum glibc requirement is only GLIBC_2.4 (from 2006). cflinuxfs5's glibc 2.39 is fully compatible.
- `sealights-agent` - add to dependency config and custom dependency watcher. SeaLights Java agent is a pure JAR-based zip, no native binaries, no glibc dependency thus no issues on cflinuxfs5.
- `datadog-javaagent` - add to dependency config, `maven` watcher is used. No native binaries, no glibc dependency with regard to cflinuxfs5.
- added also tomcat 9 to automatic updates

### DIFF
--- a/dockerfiles/depwatcher-go/internal/factory/factory.go
+++ b/dockerfiles/depwatcher-go/internal/factory/factory.go
@@ -336,6 +336,14 @@ func CheckWithClient(source Source, currentVersion *base.Internal, client base.H
 		watcher := watchers.NewGroovyWatcher(client)
 		versions, err = watcher.Check()
 
+	case "sealights_agent":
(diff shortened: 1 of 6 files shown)


ramonskie (dismissed):

kiril-keranov:
@ramonskie thx! Please merge it also

ivanovac (approved):
lgtm

PR #642, opened 2026-05-26
Return is_null

kiril-keranov:
Return `is_null` method wrongly removed. Linked to https://github.com/cloudfoundry/buildpacks-ci/pull/641

### DIFF
--- a/tasks/update-buildpack-dependency/run.rb
+++ b/tasks/update-buildpack-dependency/run.rb
@@ -23,7 +23,7 @@ end
 def sort_versions(versions)
   versions.sort_by do |v|
     if date_version?(v)
-      v
+      [v]
     else
       parsed = SemVer.parse(v.to_s) || SemVer.parse("#{v}.0")
       parsed ? [parsed.major, parsed.minor, parsed.patch] : [0, 0, 0, v]
--- a/tasks/update-buildpack-dependency/run.rb
+++ b/tasks/update-buildpack-dependency/run.rb
@@ -26,7 +26,7 @@ def sort_versions(versions)
       v
     else
       parsed = SemVer.parse(v.to_s) || SemVer.parse("#{v}.0")
-      parsed ? [parsed.major, parsed.minor, parsed.patch] : v
+      parsed ? [parsed.major, parsed.minor, parsed.patch] : [0, 0, 0, v]
     end
   end
 end
--- a/tasks/update-buildpack-dependency/run.rb
+++ b/tasks/update-buildpack-dependency/run.rb
@@ -31,6 +31,10 @@ def sort_versions(versions)
   end
 end
 
+def is_null(value)
+  value.nil? || value.empty? || value.downcase == 'null'
+end
+
 config = YAML.load_file(File.join(buildpacks_ci_dir, 'pipelines/dependency-builds/config.yml'), permitted_classes: [Date, Time])
 
 BUILD_STACKS = config['build_stacks']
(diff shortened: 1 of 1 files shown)


ramonskie (commented):
## Review: Incomplete nil guard for `version_line`

The PR correctly restores `is_null` and guards line 85, but the same nil crash exists in the deprecation block that this PR doesn't address.

### Unguarded crash on line 62

`version_line` comes from:
```ruby
version_line = ENV.fetch('VERSION_LINE', nil)
```

The guard added at line 85 (`!version_line.nil?`) prevents the crash there, but line 62 calls `version_line.downcase` without protection:

```ruby
if !is_null(deprecation_date) && !is_null(deprecation_link) && version_line != 'latest'
  dependency_deprecation_date = {
    'version_line' => version_line.downcase,  # crashes if version_line is nil
```

When `VERSION_LINE` is not set, `version_line` is `nil`. `nil != 'latest'` evaluates to `true`, so the block is entered and `version_line.downcase` raises `NoMethodError`. This is the same class of bug as the one fixed at line 85.

### Suggested fix: normalize once at assignment

Rather than adding more scattered guards, normalize `version_line` at the point of assignment so all downstream uses are safe:

```ruby
raw_version_line = ENV.fetch('VERSION_LINE', nil)
version_line = is_null(raw_version_line) ? nil : raw_version_line.downcase
```

Then remove the redundant `.downcase` calls on `version_line` throughout the file, and simplify the guards:

```ruby
# line 60 - add version_line check
if !is_null(deprecation_date) && !is_null(deprecation_link) && !version_line.nil? && version_line != 'latest'

# line 85 - consistent with is_null usage elsewhere
stacks -= [stack_name] if version_line && lines.include?(version_line)
```

This also makes the `skip_lines` guard consistent with how `is_null` is used for the other ENV vars (`deprecation_date`, `deprecation_link`, `deprecation_match`).

kiril-keranov:
> ## Review: Incomplete nil guard for `version_line`
> The PR correctly restores `is_null` and guards line 85, but the same nil crash exists in the deprecation block that this PR doesn't address.
> 
> ### Unguarded crash on line 62
> `version_line` comes from:
> 
> ```ruby
> version_line = ENV.fetch('VERSION_LINE', nil)
> ```
> 
> The guard added at line 85 (`!version_line.nil?`) prevents the crash there, but line 62 calls `version_line.downcase` without protection:
> 
> ```ruby
> if !is_null(deprecation_date) && !is_null(deprecation_link) && version_line != 'latest'
>   dependency_deprecation_date = {
>     'version_line' => version_line.downcase,  # crashes if version_line is nil
> ```
> 
> When `VERSION_LINE` is not set, `version_line` is `nil`. `nil != 'latest'` evaluates to `true`, so the block is entered and `version_line.downcase` raises `NoMethodError`. This is the same class of bug as the one fixed at line 85.
> 
> ### Suggested fix: normalize once at assignment
> Rather than adding more scattered guards, normalize `version_line` at the point of assignment so all downstream uses are safe:
> 
> ```ruby
> raw_version_line = ENV.fetch('VERSION_LINE', nil)
> version_line = is_null(raw_version_line) ? nil : raw_version_line.downcase
> ```
> 
> Then remove the redundant `.downcase` calls on `version_line` throughout the file, and simplify the guards:
> 
> ```ruby
> # line 60 - add version_line check
> if !is_null(deprecation_date) && !is_null(deprecation_link) && !version_line.nil? && version_line != 'latest'
> 
> # line 85 - consistent with is_null usage elsewhere
> stacks -= [stack_name] if version_line && lines.include?(version_line)
> ```
> 
> This also makes the `skip_lines` guard consistent with how `is_null` is used for the other ENV vars (`deprecation_date`, `deprecation_link`, `deprecation_match`).

Addressed the comments

ivanovac (approved):
lgtm

PR #644, opened 2026-05-28
Fix race condition in multi-stack dependency builds

ramonskie:
## Problem

For dependencies built against multiple stacks (e.g. httpd with cflinuxfs4 + cflinuxfs5), each stack's build ran in `in_parallel` and independently pushed to the `builds` git repo with `rebase: true`. The `update-*` job triggers on the `builds` resource and fired as soon as the **first** stack's push landed — before the second stack had committed its metadata.

This caused incomplete manifest updates: `default_versions` bumped to the new version but only one stack's dependency entry added, leaving the other stack on the old version. The `buildpack-packager` then fails with `No matching default dependency <dep> for stack <stack>`.

## Solution

- Add a `tasks/merge-builds-metadata` task that seeds from the `builds` git resource, copies all per-stack JSON files in, and makes **one atomic commit** covering all stacks.
- In the pipeline, multi-stack build tasks now run with `SKIP_COMMIT=true` (writing the JSON file but not committing). S3 uploads remain parallel and unchanged.
- After the `in_parallel` block, a single `merge-builds-metadata` task + `put: builds` replaces all per-stack puts.
- Single-stack deps (`any_stack: true`) are unaffected — they keep the existing behaviour with `SKIP_COMMIT=false` and a direct put.

## Files changed

- `pipelines/dependency-builds/pipeline.yml` — restructured build job step sequence
- `tasks/merge-builds-metadata/task.yml` — new task definition
- `tasks/merge-builds-metadata/merge.sh` — new merge script

---

## Follow-up: stack-agnostic manifest update fix

A second race condition was identified and fixed in `Dependencies#switch`: when a stack-specific dep (e.g. node, python) already exists in the manifest for one stack and a second stack's build arrives later, the update job would either silently no-op (`SKIP: Built version is not required by buildpack`) or incorrectly remove the first stack's entry.

### Root cause

`same_dependency_line?` uses a stack-subset check, so when the incoming dep has a **disjoint** stack set from the existing manifest entry, `@matching_deps` is empty. `latest?` then returns `true` vacuously, taking the "newer version" branch and replacing the existing entry instead of appending. If `latest?` returned `false`, it was a silent no-op.

Additionally, `run.rb` exited early when `added` was empty — even if the manifest changed because a new stack entry was appended.

### Fix

- **`dependencies.rb`**: Added a new branch in `switch` — before `keep_all`/`latest?` — that detects "version exists under a disjoint stack" and appends the new per-stack entry without touching existing entries. Added private method `version_exists_for_different_stack?`. Fix is fully stack-agnostic (no hardcoded stack names).
- **`run.rb`**: Captures stacks before/after `Dependencies#switch`. If stacks changed but `added` is empty, sets `rebuilt[-1] = true` to bypass the early-exit guard and ensure the manifest is committed.
- **`dependencies_spec.rb`**: Added tests for the race condition scenario and idempotency (re-processing an already-present stack entry produces no duplicate).


tnikolova82 (commented):
## Bug in `merge.sh`: count logic causes early exit — script never commits

### Problem

In `tasks/merge-builds-metadata/merge.sh`, the dry-run rsync (lines 23-26) runs **after** the real rsync (lines 19-21). Since both use `--ignore-existing` and the destination already contains the files from the real copy, the dry-run always reports 0 files to transfer:

```bash
# Line 19-21: Real rsync — copies files successfully
rsync -a --ignore-existing \
  "${stack_dir}/binary-builds-new/" \
  "merged-builds-metadata/binary-builds-new/"

# Line 23-25: Dry-run AFTER real copy — destination already has the files,
# so --ignore-existing means nothing to do. count is always 0.
count=$(rsync -a --ignore-existing --dry-run \
  "${stack_dir}/binary-builds-new/" \
  "merged-builds-metadata/binary-builds-new/" 2>/dev/null | grep -c '\.json$' || true)
```

This means `COPIED` is always 0, which triggers the early exit:

```bash
if [[ "${COPIED}" -eq 0 ]]; then
  echo "[merge] WARNING: no per-stack JSON files found — nothing to commit"
  exit 0  # ← exits before git add/commit
fi
```

**Result: the script will never commit anything.**

### Suggested fix

Remove the count/COPIED/early-exit logic entirely (lines 14-35) and rely on the `git diff --cached --quiet` check (line 42) which already handles "nothing to commit" gracefully:

```bash
for stack_dir in *-builds-metadata; do
  [[ "${stack_dir}" == "merged-builds-metadata" ]] && continue
  [[ -d "${stack_dir}/binary-builds-new" ]] || continue

  echo "[merge] Copying JSON files from ${stack_dir}..."
  rsync -a --ignore-existing \
    "${stack_dir}/binary-builds-new/" \
    "merged-builds-metadata/binary-builds-new/"
done
```

The existing `git diff --cached --quiet` block at the end of the script already exits cleanly when there are no changes — a separate counting mechanism is redundant and in this case broken.

tnikolova82 (approved):
lgtm

PR #664, opened 2026-07-16
Migrate cflinuxfs4 pipeline to noble stemcell

plamen-bardarov:
## What

Migrates the cflinuxfs4 pipeline to a noble stemcell (jammy rootfs on a noble kernel) and fixes several deprecated Concourse resource params surfaced along the way.

## Changes

- **cf-deployment resource**: removed `tag: v21.11.0` pin so it tracks the latest (noble-era) manifest instead of the old jammy-era one.
- **gcp-stemcell resource**: switched to `bosh-google-kvm-ubuntu-noble`.
- **Smoke-test deploy**: added a `use-noble-stemcell.yml` ops-file (mirrors the existing `use-xenial-stemcell.yml`) and wired it into the `cflinuxfs4-rootfs-smoke-test-deployment` put, since that manifest (from the cflinuxfs4-release repo) hardcodes `ubuntu-jammy`.
- **cflinuxfs4-image resource**: removed the invalid `email` field (leftover from the deprecated `docker-image` resource; not valid on `registry-image`).
- **upload-to-docker job**: `registry-image` has no `import_file`/`docker import` support, so the rootfs is now converted to a single-layer OCI image explicitly:
  - `prepare-docker-context` task (formerly `rename`) also emits a `FROM scratch` / `ADD rootfs` Dockerfile, making `docker-s3` a self-contained build context.
  - added a privileged `oci-build-task` step producing `image/image.tar`.
  - the put now uses `image` + `additional_tags: version/number`, with `tag: latest` in the resource source to keep pushing `latest` (replacing the invalid `import_file`/`skip_download`/`tag_as_latest` params).

## Notes

- cflinuxfs4 remains a jammy (22.04) rootfs by design; only the stemcell/kernel moves to noble.
- The functional Docker image is unchanged (single layer = rootfs) — the conversion that the deprecated `docker-image` resource did internally is now explicit.

### DIFF
--- a/deployments/operations/use-noble-stemcell.yml
+++ b/deployments/operations/use-noble-stemcell.yml
@@ -1,0 +1,6 @@
+- path: /stemcells/alias=default
+  type: replace
+  value:
+    alias: default
+    os: ubuntu-noble
+    version: latest
--- a/pipelines/cflinuxfs4.yml
+++ b/pipelines/cflinuxfs4.yml
@@ -194,6 +194,7 @@ resources:
   type: registry-image
   source:
     repository: cloudfoundry/cflinuxfs4
+    tag: latest
     username: ((cfbuildpacks-dockerhub-user.username))
     password: ((cfbuildpacks-dockerhub-user.password))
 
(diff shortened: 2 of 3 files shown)


tnikolova82 (approved):
lgtm

PR #665, opened 2026-07-17
Add pipeline architecture documentation

tnikolova82:
## Summary

Add comprehensive documentation explaining how the buildpacks-ci pipelines work, their architecture, workflows, and maintenance procedures.

## New Files

1. **`docs/pipeline-architecture.md`** (515 lines)
   - High-level overview of all buildpacks-ci pipelines
   - Pipeline relationships and dependency flows
   - Quick reference table and architecture diagrams
   - Maintenance tasks and troubleshooting

2. **`docs/buildpack-release-pipeline-overview.md`** (540 lines)
   - Detailed guide to the buildpack release pipeline
   - Two-tier testing strategy: Docker-based switchblade tests + real CF integration tests
   - Workflow from PR merge to GitHub release
   - Artifact building, version management, and ship-it process
   - Common operations and troubleshooting scenarios

3. **`docs/dependency-pipeline-overview.md`** (378 lines)
   - Complete guide to the dependency-builds pipeline
   - Dependency monitoring, building, and updating workflows
   - Stack-specific builds and multi-stack support
   - Common operations for adding/updating dependencies

### pipeline-architecture.md
- Quick reference table of all pipelines
- High-level ASCII diagrams showing pipeline relationships
- Individual pipeline details (dependency-builds, buildpack pipelines, cf-release, brats, verification)
- BOSH environments, artifacts, repositories, and maintenance tasks

### buildpack-release-pipeline-overview.md
- Simple workflow diagram
- Key concepts: pipeline jobs, artifacts, version management, stacks
- Two types of integration tests (Docker-based vs. real CF)
- Typical workflows: daily development and manual releases
- Testing strategy and triggering mechanisms
- Common operations and troubleshooting

### dependency-pipeline-overview.md
- What is the dependency-builds pipeline
- How it works: detection, building, and updating
- Pipeline configuration and per-language values
- Typical workflows for automatic and manual updates
- Common operations: adding dependencies, handling EOL versions
- Troubleshooting scenarios

## Target Audience

Designed with a layered approach:
- **Quick reference** at top for experienced users
- **Detailed explanations** with diagrams for new team members
- **Practical maintenance tasks** for operators
- **Troubleshooting guides** for common issues

## Notes

- Excludes cflinuxfs4/cflinuxfs5 stack pipelines (separate concern)
- Cross-references existing README.md for pipeline update commands
- Includes ASCII diagrams and workflow visualizations
- Documents actual triggering behavior (not periodic, but auto-triggered based on test completion)

PR #666, opened 2026-07-17
Add java-buildpack cf-release release guide

stokpop:
Adds `docs/cf-release-buildpack-guide.md`: how to cut a buildpack release
through the `cf-release` pipeline, with java-buildpack 5.0.6 as the worked
example. Written for a first-timer, via the Concourse UI rather than `fly`.

Covers three things that are easy to get confused:

- **`ship-it` is shared across all buildpacks.** One shared trigger file, so
  any buildpack whose cats-passed RC is ahead of its published version
  publishes too -- not only yours. Includes a read-only way to check first.
- **`publish` is not a second button.** It fires automatically after
  `ship-it`. A missing publish nearly always means `ship-it` was not pressed.
- **`create-cf-infrastructure-and-execute-integration-test-for-java` is not a
  gate.** It runs 5-12h on a parallel branch, so the release does not wait on
  it; the guide documents it as a manual check. The switchblade specs do gate
  the release, via `detect-new-version-and-upload-artifacts`.

Complements #665, which documents the buildpack pipeline but not
`cf-release`, CATS, or the shared trigger. The reference to
`buildpack-release-pipeline-overview.md` is plain text until #665 merges.

Job names, timings and gating checked against the live pipelines.

### DIFF
--- a/docs/cf-release-buildpack-guide.md
+++ b/docs/cf-release-buildpack-guide.md
@@ -1,0 +1,215 @@
+# Releasing java-buildpack to Cloud Foundry
+
+This guide walks through cutting a java-buildpack release and getting it into a
+Cloud Foundry BOSH release, using 5.0.6 as the worked example. It is aimed at
+someone doing this for the first time. Everything here is done by clicking in the
+Concourse web UI.
+
+The focus is the [cf-release](../pipelines/cf-release) pipeline -- the part that
+wraps a shipped buildpack into a BOSH release, tests it, and publishes it. The
+`java-buildpack` pipeline that produces the buildpack itself is only summarised
+here; see `buildpack-release-pipeline-overview.md` for that side in detail.
+
+Other buildpacks follow the same process; substitute the name throughout.
+
+## Overview
+
+A release crosses two pipelines:
+
+1. **`java-buildpack`** builds and ships the buildpack, producing a GitHub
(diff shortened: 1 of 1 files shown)


wayneeseguin (approved):
I've read through it, some things are surprising to me like all of the build packs being tied together for shipping. Other than that surprise it looks good to me.

ivanovac (dismissed):
lgtm

docs/cf-release-buildpack-guide.md
```diff
@@ -0,0 +1,215 @@
...
+here; see `buildpack-release-pipeline-overview.md` for that side in detail.
+
+Other buildpacks follow the same process; substitute the name throughout.
+
+## Overview
+
+A release crosses two pipelines:
+
+1. **`java-buildpack`** builds and ships the buildpack, producing a GitHub
+   release tagged `v5.0.6`.
```
kiril-keranov, line 20: can be pointed that `v5.0.6` is an example release version

kiril-keranov (dismissed):
looks good, just minor comment

docs/cf-release-buildpack-guide.md
stokpop: fixed

kiril-keranov (approved):
looks good

PR #667, opened 2026-07-17
Fix receipt-diff resource path so bbl-up triggers

plamen-bardarov:
The receipt-diff git resource filtered on paths: [receipt.cflinuxfs4.x86_64], a file that has never existed in public-buildpacks-ci-robots. The receipts actually live at receipt-diffs/cflinuxfs4-diff.

This was silently harmless on older git-resource versions, which ignored paths when tag_filter was set. Newer versions intersect paths with the tag_filter, so the nonexistent path filtered out every newpackages_cflinuxfs4_* tag and bbl-up stopped triggering. Point paths at the real file.


ivanovac (approved):
lgtm

PR #669, opened 2026-08-27
Fix yourkit dependency watcher

kiril-keranov:
---

Currently `build-your-kit-profiler-latest` job would run forever if triggered to seek new your-kit-profiler version.

Root cause

The YourKit download page (yourkit.com/download/) changed its structure. It no longer lists direct .zip file links — it now uses installer redirect links (/dl/installer?product=yjp&arch=x86-64&os=Linux&version=2026.3). The old regex matched zero links, so Check() returned an empty list, causing the Concourse resource to poll forever finding nothing new.

Additionally the current version (2026.3) uses a new filename format: YourKit-Java-Profiler-2026.3.176-x64.zip (no -b<build> separator, different hostname, full version in path).

Fix (yourkit.go)

Check() now scrapes two sources:
1. https://www.yourkit.com/download/archive.jsp — still has old-style zip links for 2025.9, 2025.3, etc. (regex unchanged)
2. https://www.yourkit.com/download/ — extracts the current version from the page text "Version: 2026.3" 

In() now tries the new-style URL first (download.yourkit.com/yjp/YEAR.MINOR.PATCH/YourKit-Java-Profiler-...) and falls back to the old archive URL (archive.yourkit.com/yjp/YEAR.MINOR/YourKit-JavaProfiler-...-b<build>-x64.zip) if the new URL returns non-200.

### DIFF
--- a/dockerfiles/depwatcher-go/pkg/watchers/yourkit.go
+++ b/dockerfiles/depwatcher-go/pkg/watchers/yourkit.go
@@ -18,36 +18,17 @@ func NewYourKitWatcher(client base.HTTPClient) *YourKitWatcher {
 }
 
 func (w *YourKitWatcher) Check() ([]base.Internal, error) {
-	resp, err := w.client.Get("https://www.yourkit.com/download/")
-	if err != nil {
-		return nil, fmt.Errorf("fetching downloads page: %w", err)
-	}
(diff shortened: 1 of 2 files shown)


ivanovac (approved):
lgtm

tnikolova82 (approved):

PR #672, opened 2026-09-17
Update create-new-version-line-story.rb

kiril-keranov:
---

**Fix `SemanticVersion` to handle major-only and 4-part version strings**

**Problem**

The `create-new-version-line-story` Concourse task was failing for SAPMachine with:
```
ArgumentError: Not a semantic version: "27"
```

SAPMachine publishes two types of version strings that the old regex couldn't handle:
- Major-only (e.g. `27`) — emitted when a new LTS major line is released, after stripping the `sapmachine`- tag prefix
- 4-part versions (e.g. `17.0.20.1, 21.0.12.1, 25.0.4.1`) — SAPMachine's standard patch release format

The previous regex required at minimum major.minor, so "27" raised an `ArgumentError` before the existing-line filter could even run.

**Fix**

Updated `SemanticVersion`#initialize to accept 1–4 numeric segments, with minor, patch, and the new build segment all defaulting to 0 when absent:

^v?(\d+)(?:\.(\d+)(?:\.(\d+)(?:\.(\d+))?)?)?(.+)?$

- `"27"` → `major=27, minor=0, patch=0, build=0` — no existing line matches, dispatch fires correctly
- `"17.0.20.1"` → `major=17, minor=0, patch=20, build=1` — `SemanticVersionFilter` compares major.minor.patch = "17.0.20", which matches filter 17.X.X
- All existing 2/3-part version strings (e.g. "1.27.3", "3.10.5") parse identically to before

**Impact**

No other dependencies use major-only or 4-part versions — all other components in `config.yml` use standard `major.minor.X` or `major.X.X` patterns. Fully backwards-compatible.


stokpop:
LGTM (side question: so these 4 way versions are working correctly in the java-build-pack itself?)

stokpop (approved):

kiril-keranov:
> LGTM (side question: so these 4 way versions are working correctly in the java-build-pack itself?)

It appears some changes will be needed in the buildpack also for handling them. I might merge this in order to fix the failing `create-sapmachine-new-line-story` for sapmachine and further address the java buildpack stuff

PR #675, opened 2026-09-24
Add ruby buildpack as a yarn-berry consumer

tnikolova82:
## Summary
Adds a `ruby` block to the `yarn-berry` dependency entry in `pipelines/dependency-builds/config.yml`, mirroring the existing classic `yarn` entry (which already lists both `nodejs` and `ruby`).

## Why
This renders a new `update-yarn-berry-4.x.x-ruby` job, which will open an automated manifest.yml PR against `ruby-buildpack` - the same mechanism that already produces `update-yarn-latest-ruby` for classic `yarn` (https://concourse.app-runtime-interfaces.ci.cloudfoundry.org/teams/buildpacks-team/pipelines/dependency-builds/jobs/update-yarn-latest-ruby/builds/7).

## Scope note
This only adds the dependency to `ruby-buildpack`'s manifest via an auto-generated PR - it does not make `ruby-buildpack` actually use Yarn Berry. `ruby-buildpack`'s `InstallYarn()` is currently hardcoded to `Installer.InstallOnlyVersion("yarn", ...)` with no Yarn Berry detection logic (same situation as `nodejs-buildpack` before this). That follow-up supply-logic work is being tracked/implemented separately for both buildpacks.

## Test plan
- `pipelines/dependency-builds/config.yml` validated as well-formed YAML.


ivanovac (approved):
lgtm

